PR #41, opened 2014-11-24
Issue 31: scores.csv file needs to have a bit more information

### LINKED ISSUE #31: scores.csv file needs to have a bit more information
After running NAB you get a _scores.csv file with some useful information. This file should contain most (if not all) the information needed to manually calculate the score. For example, the fn column contains the total number of records. Instead it should contain the number of anomaly windows that were unlabeled because that is what the score is based on.

simjega:
Fixes #31

This adds a summary line for each score.csv file
This also makes each score.csv file specific to a given combination of detector, profile, and threshold.

### DIFF
--- a/nab/optimizer.py
+++ b/nab/optimizer.py
@@ -136,11 +136,8 @@ def objectiveFunction(threshold, args):
   if not 0 <= threshold <= 1:
     return float("-inf")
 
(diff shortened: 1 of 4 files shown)


BoltzmannBrain:
Closing this b/c replaced by PR#56.

Issue #61, opened 2015-01-19
For combining labels, need a buffer for anomaly start times 


BoltzmannBrain:
Anomalies that differ by a few time steps should not be thrown away when combining labels, but a requirement is then a ground truth start time must be determined.

PR #62, opened 2015-01-19
Standardized relaxed window sizes

### LINKED ISSUE #59: Relaxed windows overlapping
We need a means of both defining anomaly windows and calculating the subsequent relaxed windows such that the relaxed windows do not overlap.

BoltzmannBrain:
Fixes #59 
New combined_labels.json reflects the new relaxed window calculations.
Some debugging code remains (commented out) and will be removed later.


BoltzmannBrain:
@subutai please review; fixes #59 and #62.

The revised label combining process here is:
For each timestamp in all datafiles, the timestamp is the start of an anomaly if all three (>90%) labelers agree it is. With the total number of anomalies in each datafile, the respective relaxed window lengths are calculated. For each anomaly start time, the relaxed windows are placed such that 1/2 of the window precedes the timestamp (and 1/2 follows). The central positioning maintains the scoring scheme where an identified anomaly will score 0.98661, 0.84828, and 0.0 if located at the left edge of a window, middle, and right edge, respectively. Outside the window (false positives) yields negative values.

Fixing issue #61 is a followup to this PR.

BoltzmannBrain:
@subutai the more I think on it the more I'm convinced the "return to normal" (i.e. the end of an anomaly window) is an important part of scoring and shouldn't be a rough estimate (= 10% of data length divided by the number of anomalies). That is, we may want to ignore this PR and implement a different method.

subutai:
@BoltzmannBrain Given the current data and labels, would it make a big difference if we made it a more precise estimate?

BoltzmannBrain:
@subutai been looking into that today. Still tinkering with and coding up some different approaches for both relaxing the windows and combining the labels. For now it looks like our best bet is to make a precise estimate for anomaly length, and use that along with the hand-labeled start times.

BoltzmannBrain:
@subutai please review. This fixes issues #59 and #61

nab/labeler.py
```diff
@@ -188,15 +192,14 @@ def write(self, destPath):
...
     self.getUserLabels()
-    self.combineLabels()
-    self.combineWindows()
+    self.combineRawLabels()
     self.relaxWindows()
 
 
   def getUserLabels(self):
-    """Collect UserLabels."""
+    """Collect the raw user labels from the path 'nab/labels/raw/'."""
```
subutai: The path is not hardcoded is it?

nab/labeler.py
```diff
@@ -206,111 +209,104 @@ def getUserLabels(self):
...
+        else:
+          rawAnomalies.append(bucket)
+          current = t
+          bucket = [current]
+      if bucket:
+        rawAnomalies.append(bucket)
+
+      # Merge the bucketed timestamps that qualify as true anomalies
+      for bucket in rawAnomalies:
+        if len(bucket) > len(self.userLabels)*self.threshold:
```
subutai: Should it be `>=` here? If threshold is 50% then we would want 2 out of 4 to pass.

nab/labeler.py
BoltzmannBrain: It looks in the directory for "labels/raw". This is in scripts/combine_labels.py line 81. This is the organization for label files we want, no (issue#57)?

nab/labeler.py
BoltzmannBrain: You're right, I'll fix this.

nab/labeler.py
```diff
@@ -206,111 +209,104 @@ def getUserLabels(self):
...
+    for relativePath, anomalies in self.labelIndices.iteritems():
+    
       data = self.corpus.dataFiles[relativePath].data
-      length = len(data["timestamp"])
-
-      relaxWindowLength = int(percentOfDataSet*length)
+      length = len(data)
+      num = len(anomalies)
+      if num:
+        relaxWindowLength = int(self.windowSize * length / len(anomalies))
```
subutai: is windowSize a float? length and anomalies are ints, so just worried about integer truncation due to the divide.

subutai:
@BoltzmannBrain Left some comments. Looks good otherwise. Do we have any tests of the label combiner?

nab/labeler.py
subutai: Right, that's what we want for NAB. But in this class it is not a hardcoded path. getUserLabels looks in self.labelDir

nab/labeler.py
BoltzmannBrain: Yes windowSize is a float, but is this a problem? The calculation rounds down to the nearest int, which I think is okay and desired here. This would be a problem for rounding down to 0, but not an issue for us b/c only would happen with really small dataset lengths.

nab/labeler.py
BoltzmannBrain: Okay got it, will revise.

BoltzmannBrain:
Yes, corpuslabel_test.py. All tests pass :)

nab/labeler.py
subutai: OK. I don't think it's a problem.

subutai:
:+1:

Issue #68, opened 2015-01-24
Better command line UX


BoltzmannBrain:
Users should be able to easily set parameters, particularly which individual detectors and profiles to run because they take a while. They can do this manually in the code, but it would be much better from command line (with supporting instructions and examples in the readme).

subutai:
How is this different from existing options such as `profilesFile`?

BoltzmannBrain:
I was trying to think of how NAB can be simpler (more accessible) to new users. You're right, I was not aware the user could change the arguments from the command line. However it's not explicitly stated anywhere how to do this. Do you think we need to explain these options? If so, should the details be in the readme or the writeup?

subutai:
If you do run.py --help it provides a nice description of all the options. This is similar to most Unix commands. We can put that note in the README.

PR #70, opened 2015-01-25
readme updates

BoltzmannBrain:
@subutai please review.
Added the option for the user to select which detector(s) to run. May fix issue #68 but I'm unsure if you think the user should have the option to easily change more parameters or not.
Also edited the readme to reflect this change, as well as some instructions on how to test your own detector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,14 @@
 ### Under Construction
 
-#### This repo is under heavy, active revision. The code does not work yet. Browse, but don't judge. :)
+#### This repo is under active revision. The code runs but is not yet complete, so please don't judge. :)
 
 The Numenta Anomaly Benchmark
 -----------------------------
 
-This repository contains the data and scripts necessary to replicate the
-results in the forthcoming Numenta Anomaly Benchmark (NAB) paper.
+Welcome. This repository contains the data and scripts necessary to replicate the
+results in the forthcoming Numenta Anomaly Benchmark (NAB) paper. It also provides the tools to run NAB to score your own anomaly detection algorithms. Competitive results tied to open source code will be posted here on the Scoreboard. Let us know about your work by submitting a pull request.
 
-We hope you will compare these results with your own anomaly detection methods.
-Competitive results tied to open source code will be posted here on the Scoreboard. Let us know
-about your work by submitting a pull request.
-
-### Corpus
+#### Corpus
 
 The NAB corpus of timeseries datasets is designed to provide data for research
 in streaming anomaly detection. It is comprised of both artificial and
@@ -57,10 +53,7 @@ You need to manually install the following:
 - [Python 2.7](https://www.python.org/download/)
 - [pip](https://pip.pypa.io/en/latest/installing.html)
 - [Numpy](http://www.numpy.org/num)
-- [NuPIC](http://www.github.com/numenta/nupic) (Source Required)
-
-It is also assumed you have a full checkout and working version of NuPIC 
-by this point.
+- [NuPIC](http://www.github.com/numenta/nupic) (Required only to run the Numenta detector)
 
 ##### Download this repository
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,18 +1,14 @@
 ### Under Construction
 
-#### This repo is under heavy, active revision. The code does not work yet. Browse, but don't judge. :)
+#### This repo is under active revision. The code runs but is not yet complete, so please don't judge. :)
```
subutai: I would prefer to say "does not work yet" until we feel confident it actually does work well.

README.md
```diff
@@ -72,16 +65,28 @@ by this point.
...
-This will install the requirements, such as pandas and simplejson.  
+This will install the additional required modules pandas and simplejson.  
 
 
 ### Usage
 
+##### Run NAB
+
     cd /path/to/nab
-    python setup.py develop (for now)
```
subutai: Isn't this line still needed?

README.md
```diff
@@ -72,16 +65,28 @@ by this point.
...
     python run.py
 
-This will produce results files for the Numenta anomaly detection method, as well as baseline results using methods from the [Etsy Skyline](https://github.com/etsy/skyline) anomaly detection library, and a random detector. This will also pass those results files to the analyze_results.py script to generate final scores.
+This will produce results files for the desired anomaly detection methods; the user is prompted which algorithm(s) to run. Included in the repo are the Numenta anomaly detection method, as well as methods from the [Etsy Skyline](https://github.com/etsy/skyline) anomaly detection library, and a random detector. This will also pass those results files to the analyze_results.py script to generate final NAB scores.
+
+##### Run your detector
+
+To run your own detector (named 'DUT' for detector under test), first create a folder in the nab detectors directory:
```
subutai: This statement is actually misleading as there are a couple of different ways to run your detector including a variation that is much simpler than what is specified here. I think we should take the details out of the README and instead point them to the Word document.

subutai:
@BoltzmannBrain I'm confused by this PR. It contains a whole bunch of changes not related to the issue, such as changes to the README. In addition I'm not sure what you mean by #68. The user can already choose detectors, profiles, etc. through the command line options.

BoltzmannBrain:
@subutai the changes to the readme were to support the change to run.py that prompts the user to select the detector. Sorry for the confusion.
I agree we should stick to using the command line options -- please see my comment on issue #68.

subutai:
Well, there were a bunch of other unrelated changes to the text (see my inline comments).

README.md
BoltzmannBrain: Yes you're right. Do you know who put the `(for now)` part, and what we intend to use instead?

BoltzmannBrain:
This PR now just looks at revising the readme (leaving run.py as is).
Currently the NAB writeup points to the readme for instructions on how to add a detector. Do you think the instructions would be best in the writeup or the readme?

README.md
subutai: I put that in there - it's required for developing on the codebase, so mostly for us. Once we feel more comfortable about it, we could take out the `develop` portion.

subutai:
> Do you think the instructions would be best in the writeup or the readme?

There are three different ways to use a new detector with NAB. The most complex (and therefore least likely to be used) is adding a new detector that conforms to the detector API. Wherever we put it, we need to discuss all three options and the tradeoffs - seems like the writeup is the best place for that.  The README should be very short.

BoltzmannBrain:
Sure, I agree the readme should be concise. On that note, do you think we'd want to move the "Data format" section to the writeup as well?

What are the three ways to use a different detector?

README.md
BoltzmannBrain: Got it, thank you for the clarification :)

subutai:
> do you think we'd want to move the "Data format" section to the writeup as well?

Hmm, that's not a bad idea. We could just say the data formats are very simple - please see the writeup for more details.

> What are the three ways to use a different detector?

1. Conform to the API. This involves writing NAB specific Python code.
2. Output anomaly scores to a set of conforming files. Then run our optimizer to threshold those files and then run the scoring function. This is much easier than 1. 
3. Output thresholded files and just run our scoring function. 

run.py was structured to allow any of the above paths - hopefully it still is! I just looked through the last version of the writeup - I don't see the above three described. There used to be a really clear section (with diagram) that showed the three options. I wonder where that is! Let's discuss in our next meeting.

subutai:
:+1:

Issue #80, opened 2015-02-04
Identify raw labels that don't make it into combined labels


BoltzmannBrain:
For debugging the labeler, it would be helpful to print out the raw labels that do not qualify (beat the threshold) for the combined labels.

Issue #78, opened 2015-02-04
Merge overlapping windows


BoltzmannBrain:
If combined labels overlap, merge them into one anomaly. Justification is (i) multiple anomalies don't occur in the same window, and (ii) although anomalies close together in time may appear distinct, they're likely(?) correlated.

Issue #77, opened 2015-02-04
Incorporate larger windows


BoltzmannBrain:
Larger window sizes (>10%) are preferred, and windows centered about the true anomaly. This is to reward early detection, even if humans can't see the anomaly visually, and still reward late detection b/c it's better to identify an anomaly late rather than not at all.

BoltzmannBrain:
PR#82 increased the window size to 10% of datafile length, but this issue remains open b/c we may want to increase this further. Downgrading to P3, however.

BoltzmannBrain:
We've settled on 10% window size (for v0.1, private beta).

PR #82, opened 2015-02-04
revised window calcs and label combining; issues #77 #78 #80

BoltzmannBrain:
@subutai please review
- issue #77 : window amount (for each data file) increased to 10% from 7%
- issue #78 : overlapping windows are now merged into one larger window, taking the first and last timestamps of the union
- issue #80 : the raw labels that don't qualify for the combined labels set (i.e. their bucket size does not exceed the threshold) are printed to the console (along with the respective filename)
- additional: 
    - added `verbosity` variable for the user to set the amount of labeling metrics printed out
    - combined_labels.json is the new set of ground truth labels
    - `addKnownLabels` method at the end of labeler.py is for manually adding labels


nab/labeler.py
```diff
@@ -266,14 +276,22 @@ def combineRawLabels(self):
 
       # Merge the bucketed timestamps that qualify as true anomalies
       for bucket in rawAnomalies:
+#        import pdb; pdb.set_trace()
```
subutai: Should get rid of this in committed code.

nab/labeler.py
```diff
@@ -312,3 +336,42 @@ def relaxWindows(self):
...
+#      print self.combinedRelaxedWindows
+      num_windows = len(windows)
+      if num_windows > 1:
+        for i in range(num_windows-1):
+          if (pandas.to_datetime(windows[i+1][0])
+              - pandas.to_datetime(windows[i][1])).total_seconds() <= 0:
+            windows[i] = [windows[i][0], windows[i+1][1]]
+            del windows[i+1]
+#            import pdb; pdb.set_trace()
+#      print "------2-------"
```
subutai: Should these print statements be conditioned on self.verbosity?

nab/labeler.py
```diff
@@ -312,3 +336,42 @@ def relaxWindows(self):
...
+#      print relativePath
+#      print self.combinedRelaxedWindows
+#      self.combinedRelaxedWindows[relativePath] = windows
+#      print "------3-------"
+#      print relativePath
+#      print self.combinedRelaxedWindows
+#    import pdb; pdb.set_trace()
```
subutai: Get rid of import pdb statements in committed code.

subutai:
Left a few comments. :+1: otherwise.

BoltzmannBrain:
Sorry, I really thought I had removed those! Thanks @subutai

PR #85, opened 2015-02-05
label bucketing buffer now considers the 5-minute granularity

BoltzmannBrain:
@subutai please review
Anomaly start timestamps were not bucketing properly; timestamps that should have been bucketed together were not. The issue was the buffer size calculation is in minutes, but the data files are at a granularity of 5-minutes. The included labels/combined_labels.json file is the result of the label combining process after this fix.


subutai:
:+1:

Issue #79, opened 2015-02-04
Adjust labels for "known cause" data files


BoltzmannBrain:
For the data files which we know the anomaly cause(s, we should manually label the windows. This manual labeling should also be noted in the writeup.

PR #86, opened 2015-02-07
revised truth label process for new windows and known labels

BoltzmannBrain:
@subutai please review, some big changes to the label combining process:
- nab/labeler.py is revised to accommodate the new raw labels (i.e. only start timestamps, not windows) and known labels, and functions are cleaned up a bit
- added the known labels as labels/raw/known_labels.json
- updated the raw human labels to reflect anomaly start times, not windows
- updated labels/combined_labels.json per the above changes

**Note**: I'm not completely satisfied with labeler.py yet and plan to modify it a bit more. But changes are tough because of other dependencies throughout NAB, so I wanted to get your feedback on this PR first :)


BoltzmannBrain:
fixes issue #79

nab/labeler.py
```diff
@@ -54,13 +54,15 @@ def __init__(self, path, corpus):
 
     self.corpus = corpus
     self.getWindows()
-    self.getLabels()
+    
+    if "raw" not in self.path:
```
subutai: We should mention in the comments the special substring "raw" in path.

nab/labeler.py
```diff
@@ -76,31 +78,43 @@ def found(t, data):
...
-      if not all(deltas):
-        raise ValueError("In the label file %s, timestamps are not in "
-                         "chronological order." % self.path)
-      
+
+
+  def validateLabels(self):
+    """
+    This is run at the end of the label combining process (see
+    scripts/compine_labels.py) to validate the resulting ground truth windows.
```
subutai: spelling: `combine_labels`.  Also, what is it validating?

nab/labeler.py
```diff
@@ -194,52 +208,62 @@ def write(self, destPath):
...
+    is merged into one timestamp -- the ground truth label.
+    The set of known anomaly labels are added as well.
     
     If verbosity>0, the dictionary passedLabels -- the raw labels that did not
     pass the threshold qualification -- is printed to the console.
     """
-    combinedLabels = {}
-    labelIndices = {}
-    passedLabels = {}
+    def setTruthLabels(dataSet, trueAnomalies):
```
subutai: Comment explaining this function. Also, any particular reason to define inside method?

nab/labeler.py
```diff
@@ -194,52 +208,62 @@ def write(self, destPath):
...
-    if len(self.userLabels) == 0:
+    self.nLabelers = len(self.userLabels)
+    if self.nLabelers == 0:
       raise ValueError("No users labels found")
 
-    self.nlabelers = len(self.userLabels)
-
     
-  def combineRawLabels(self):
+  def combineLabels(self):
```
subutai: This method is getting pretty long and hard to follow. I would recommend breaking into smaller parts, and documenting each part really clearly.

nab/labeler.py
```diff
@@ -194,52 +208,62 @@ def write(self, destPath):
...
+    Combines raw user labels to create set of true anomaly labels.
+    A buffer is used to bucket labels that identify the same anomaly. The buffer
     is half the estimated window size of an anomaly -- approximates an average
     of two anomalies per dataset, and no window can have >1 anomaly.
-    After merging, a label becomes a true anomaly if it was labeled by a
-    proportion of the users greater than the defined threshold.
+    After bucketing, a label becomes a true anomaly if it was labeled by a
+    proportion of the users greater than the defined threshold. Then the bucket
+    is merged into one timestamp -- the ground truth label.
+    The set of known anomaly labels are added as well.
```
subutai: Are the known anomaly labels treated as if every user had entered an anomaly at the same timestamp?

scripts/combine_labels.py
```diff
@@ -101,6 +106,12 @@ def main(args):
                       labelers for a record to be considered anomalous (should \
                       be a number between 0 and 1)")
   
+  parser.add_argument("--windowSize",
```
subutai: I think we should make this a constant in the code. This is not something a NAB user would be setting.

scripts/combine_labels.py
```diff
@@ -101,6 +106,12 @@ def main(args):
...
+                      default=0.10,
+                      type=float,
+                      help="Set the size of the standard anomaly window, as a \
```
subutai: This is the size of all anomaly windows combined, not the size of each window.

nab/labeler.py
BoltzmannBrain: Because it's only called in combineLabels() I thought it would be tidier, similar style to found() in getWindows(). No problem to move it outside the method though.

nab/labeler.py
BoltzmannBrain: They're added as the raw labels are combined, but they skip the bucket-merge combining process. So it is as if the users entered start timestamps for the known data files that would have resulted in them being bucketed together, surpassing the threshold, and merging into what we have set as the ground truth timestamps for these data files. I'll add to the comments to clarify, but this is also explained in the writeup.

scripts/combine_labels.py
BoltzmannBrain: Okay, this will be removed following your previous comment -- to delete the windowSize user parameter.

nab/labeler.py
subutai: OK, thanks.

nab/labeler.py
subutai: I'm fine leaving it here. Just curious.

nab/labeler.py
BoltzmannBrain: Agreed, I think I'll make separate functions for "bucketing" and "merging" timestamps.

BoltzmannBrain:
@subutai please review.
I added scripts/create_empty_label_file.py to this PR because it should not create entries for the known label files.

subutai:
:+1:

PR #90, opened 2015-02-11
renamed scoring profiles to make clearer

BoltzmannBrain:
@subutai please review; renaming the scoring profiles.

### DIFF
--- a/config/profiles.json
+++ b/config/profiles.json
@@ -7,7 +7,7 @@
     	"tnWeight": 1.0
 	}
   },
-"false_positives": {
+"reward low FP rate": {
 	"CostMatrix": {
     	"tpWeight": 1.0,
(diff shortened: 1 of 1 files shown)


config/profiles.json
```diff
@@ -7,7 +7,7 @@
     	"tnWeight": 1.0
 	}
   },
-"fp_error": {
+"false_positives": {
```
subutai: Still not clear. What do you mean by "false_positives"? It doesn't say anything about what is preferred or what is weighted higher or lower.  When looking at results, the profile name should be self explanatory without having to refer back to the profile weights each time.

BoltzmannBrain:
@subutai sorry, I was trying to keep the names short. Now the names should be clearer.

subutai:
:+1:

Issue #95, opened 2015-02-12
Scale FPs


BoltzmannBrain:
All FPs are scored against the total, yet only one TP per window is added to the total. While we are estimating x% of a data file is anomalous (i.e. the anomaly windows), and (1-x)% of the data file is normal data, then TPs are limited to x% of the data, while FPs exist in the other (1-x)%. Thus scaling the FPs by 1/x will level out the contributions made to the total score by TPs and FPs. E.g. if x=10, then the FP weight should be multiplied by 0.10.

Also add this to the writeup.

BoltzmannBrain:
For 10% window size, FPs scale to 0.11.

Issue #36, opened 2014-10-28
Label combiner should not allow relaxed windows to be in probationary period


subutai:
Label combiner should not allow relaxed windows to be in probationary period. If a window creeps into the probationary period, it should get truncated so that no part of the window goes into probationary period.

BoltzmannBrain:
Fix in PR#97 raises a ValueError so we can handle each case manually in the function `labeler.removePoorLabels()`.

PR #97, opened 2015-02-12
scaling FP weight for 0.10 window size

BoltzmannBrain:
FP weight scaled down; issue #95


BoltzmannBrain:
@subutai please review. Closes issues #95 #98 #77 #36

nab/labeler.py
```diff
@@ -349,6 +357,20 @@ def checkWindows(self):
...
             del windows[i+1]
 
+      length = len(self.corpus.dataFiles[relativePath].data)
+      probationIndex = int(math.ceil(self.probationaryPeriod*length))
+      probationTimestamp = self.corpus.dataFiles \
+                             [relativePath].data["timestamp"][probationIndex]
+      if num_windows > 0:
+        if (pandas.to_datetime(windows[0][0])
+            -probationTimestamp).total_seconds() < 0:
+#          raise ValueError("The first window in %s overlaps with the"
```
subutai: In general, shouldn't we raise an exception if this happens?

subutai:
Left one question. Other than that :+1:

nab/labeler.py
BoltzmannBrain: Yes, but we want the code to either delete the bad window, or remove the file completely, and then keep running. Are you saying the ValueError should be raised in a try-except block?

subutai: No, I mean an exception. In general I don't think we should proceed without manual intervention in the case where the probationary period overlaps the first anomaly window.

BoltzmannBrain:
Okay, so the ValueError exception is raised for the data file "realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_NetworkIn.csv". Should we remove it from the benchmark dataset? Removing the window would result in FPs.

BoltzmannBrain:
@subutai please review. As we add more data files and other instances of poor labels arise, we can remove them in the same function -- `removePoorLabels()`.

Issue #96, opened 2015-02-12
NAB versioning


BoltzmannBrain:
Implement a versioning system for NAB. This will regulate NAB updates including additions to the benchmark dataset, the scoreboard, and any code changes.

BoltzmannBrain:
First version for private beta is v0.1.

Issue #75, opened 2015-01-29
Test repo installation


BoltzmannBrain:
Need to test new installation of the NAB repo, and update readme accordingly.

BoltzmannBrain:
Following the install and run instructions in the readme works as expected.

subutai:
Great!

PR #102, opened 2015-02-17
nab setup and version number

BoltzmannBrain:
@subutai please review.
Issue #96, and #75 will be in a following commit.


subutai:
:+1:

Issue #129, opened 2015-04-16
Command line arguments that take lists should use commas


scottpurdy:
It looks like the `python run.py -d ...` option takes a space separated list of detector names. Spaces on the command line are annoying so this should be a comma-separated list instead.

PR #139, opened 2015-04-28
command line cleanup

### LINKED ISSUE #128: No spaces in filenames
I noticed that some filenames have spaces. This is annoying to work with on the command line so we should avoid it. Use underscores or something instead.

Example:

results/numenta/numenta_reward low FP rate_scores.csv

BoltzmannBrain:
Fixes #128 and #129 
@scottpurdy please review. As far as I can tell there is no way for argparse to implicitly take comma-separated input parameters. Here's the [documentation](https://docs.python.org/2.7/library/argparse.html), and my [SO question](http://stackoverflow.com/questions/29922885/python-argparse-a-list-input/29926014#29926014) explains it well.


scottpurdy:
You could try using [nargs](https://docs.python.org/3/library/argparse.html#nargs) with positional arguments. Or take a string and then do `parser.stringArg.split(",")`

PR #153, opened 2015-07-03
Boltzmann brain plotly

BoltzmannBrain:


nab/labeler.py
```diff
@@ -204,14 +204,22 @@ def __str__(self):
     return ans
 
 
-  def write(self, destPath):
-    """Write the combined labels to a destination directory."""
-    if not os.path.isdir(destPath):
-      createPath(destPath)
+  def write(self, labelsPath, windowsPath):
```
BoltzmannBrain: This should be a generic `write(filePath)` that is called twice.

results/numenta/numenta_standard_scores.csv
```diff
@@ -1,34 +1,34 @@
 Detector,Profile,File,Threshold,Score,TP,TN,FP,FN,Total_Count
```
BoltzmannBrain: These results should be the CSV from the Scoreboard run.

scripts/plot.py
```diff
@@ -0,0 +1,416 @@
...
+                   y=self.rawData['value'],
+                   name='Value',
+                   line=Line(
+                     width=1.5
+                   ),
+                   showlegend=False
+           )
+
+
+  def _createLayout(self, title):
```
BoltzmannBrain: change to staticmethod

Issue #155, opened 2015-07-03
Plotting todo


BoltzmannBrain:
Address the TODOs in scripts/plot.py.

PR #161, opened 2015-07-16
PlotNAB class improvements

BoltzmannBrain:


BoltzmannBrain:
fixes #155

BoltzmannBrain:
@scottpurdy addresses your previous review

scripts/plot.py
```diff
@@ -27,7 +27,15 @@
 import pandas as pd
 import plotly.plotly as py
 
-from plotly.graph_objs import *
+from plotly.graph_objs import (Bar,
```
scottpurdy: yikes, I'd break the first `Bar` onto a new line and then list all of the imported items on the same line, breaking only if you hit the end of the line.

alternatively, you could do `from plotly import graph_objs` and then later do `graph_objs.Bar`, etc.

scripts/plot.py
```diff
@@ -392,33 +399,33 @@ def _addDetections(self, name, symbol, FP, TP):
   # Examples:
   
   dataFiles = [
```
scottpurdy: You could make these tuples instead of lists so they are explicitly immutable.

scottpurdy:
:+1:

PR #166, opened 2015-07-31
batch of data files and labels for v1.0

BoltzmannBrain:


BoltzmannBrain:
Notes after doing my raw labels:
- "realAdExchange/exchange-4_cpm_results.csv" and "realAdExchange/exchange-4_cpc_results.csv" are almost identical... maybe remove one?
- The very end of "data/realTweets/Twitter_volume_FB.csv" could be an anomaly, but there is no more data afterwards to support an anomaly window. Should we remove these last few data instance from the file? A better sol'n may be to handle this programmatically, where if a ground truth anomaly window overlaps with the end of the file, we don't score it (or alternatively trim the file), and if it overlaps with the probationary period, we don't score it.

@tomsilver if you have any similar notes from your labeling please add them here, and then I'll create an issue to address them.

BoltzmannBrain:
@subutai the windows are ready for you to inspect, just run `python scripts/plot.py` from this PR, where you can specify which files you'd like to plot at the bottom of the script.

It currently plots the GOOG tweets file, which gives [this plot](https://plot.ly/~alavin/2471/raw-data-for-labels-inspection-for-realtweetstwitter-volume-googcsv/).

subutai:
Thanks. I was able to plot some of the new files. I will look at them more on Monday.

Issue #126, opened 2015-03-26
Add unit tests for score normalization


BoltzmannBrain:
New score normalization scheme, per #125, needs unit test coverage.

PR #168, opened 2015-08-04
created normalization unit test

tomsilver:
addressing #126 

@BoltzmannBrain please review

### DIFF
--- a/tests/unit/normalize_test.py
+++ b/tests/unit/normalize_test.py
@@ -1,0 +1,176 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


tests/unit/normalize_test.py
```diff
@@ -0,0 +1,176 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Tests normalization scheme for normalizing detectors' scores according
+to baseline."""
+
+import csv
+import json
```
BoltzmannBrain: use simplejson, as imported [here](https://github.com/numenta/NAB/blob/master/nab/labeler.py#L27-30)

tests/unit/normalize_test.py
```diff
@@ -0,0 +1,176 @@
...
+import os
+import shutil
+import tempfile
+import unittest
+
+from nab.runner import Runner
+
+
+def createTemporaryCsv(parentDir, fileName, data):
```
BoltzmannBrain: change name b/c there's nothing about this method that requires the csv (or its dir) to be temporary

tests/unit/normalize_test.py
```diff
@@ -0,0 +1,176 @@
...
+    testRunner.resultsFiles = [resultsFilePath]
+
+  return testRunner
+
+
+
+class NormalizationTest(unittest.TestCase):
+
+  def setUp(self):
+    self._tmpDirs = [] # stores absolute paths of dirs created during testing
```
BoltzmannBrain: `setUp()` and `tearDown()` are called for each test method, so this is unnecessary. If a future test creates multiple temp dirs we'll modify these methods then.

tests/unit/normalize_test.py
```diff
@@ -0,0 +1,176 @@
...
+    self.resultsHeaders = ['Detector','Profile','Score']
+
+
+  def tearDown(self):
+    for tmpDir in self._tmpDirs:
+      shutil.rmtree(tmpDir)
+
+
+  def _createTemporaryResultsDir(self):
```
BoltzmannBrain: don't need this (see above comment)

BoltzmannBrain:
@tomsilver done with review, left some comments, otherwise :+1:

tomsilver:
@BoltzmannBrain thanks, ready for re-review!

BoltzmannBrain:
:+1:

PR #177, opened 2015-09-17
Detection bugs

BoltzmannBrain:
@subutai this fixes the issues you identified by running `python run.py --detectors baseline,random`


subutai:
:+1:

PR #180, opened 2015-09-18
Misc updates

subutai:
Add path checks: These changes don't require that every data file have a window definition. This lets you to run with a small window file for debugging. It warns you if there are data files for which you don't have windows.

Updated plot example: Fixed some bugs and made the usage clearer.


subutai:
@BoltzmannBrain Let me know if this looks ok.

BoltzmannBrain:
:+1:

PR #184, opened 2015-09-23
Switch over to using the Anomaly Likelihood class in NuPIC

### LINKED ISSUE #172: Replace AnomalyLikelihood class with the new NuPIC version
The Numenta detector uses its own implementation of the AnomalyLikelihood class ([here](https://github.com/numenta/NAB/blob/4b56c59e73452e16e487467304fd54c210651d26/nab/detectors/numenta/numenta_detector.py#L108-L166)). We should look into replacing it with the [NuPIC version](https://github.com/numenta/nupic/blob/master/nupic/algorithms/anomaly_likelihood.py#L69), which has improved computation efficiency as of [#2535](https://github.com/numenta/nupic/pull/2535). Before deciding on the change it's important to run NAB to investigate the difference in speed and score.

subutai:
Changes:

Switched over to using the Anomaly Likelihood class.   Initially there was a performance drop due to some small differences in how the distribution estimation is handled.  I fixed these in NuPIC.  With those changes we get identical performance to what we see without using this class.

I also retuned the parameters a bit and improved overall scores.

I've verified that switching to log likelihood does not affect the final score, so we now output that as our anomaly score. It is much easier to plot. Interestingly the best threshold is consistently 0.5.

Fixes #172


subutai:
@BoltzmannBrain

subutai:
FYI, corresponding NuPIC pull request: https://github.com/numenta/nupic/pull/2594

BoltzmannBrain:
@subutai what are the final normalized scores you get for the Numenta detector?

subutai:
> @subutai what are the final normalized scores you get for the Numenta detector?

```
Final score for 'numenta' detector on 'reward_low_FP_rate' profile = 61.22
Final score for 'numenta' detector on 'reward_low_FN_rate' profile = 74.47
Final score for 'numenta' detector on 'standard' profile = 69.67
```

BoltzmannBrain:
:+1: 
The scoreboard should be updated to reflect these scores, correct?

subutai:
> The scoreboard should be updated to reflect these scores, correct?

Yes, but perhaps we should wait until we add in the new data files.

PR #189, opened 2015-09-24
Label combiner removes redundant raw labels

BoltzmannBrain:
For a given data file, if a user labels two records close in time to one another, the label combiner will double-count the earlier label in the bucket-merge scheme. That is, in the label combining algorithm, all raw user labels of a given data file are pooled into a list where they are bucketed together if they fall within a "time buffer" (1/2 an anomaly window) of one another. The buckets are "merged" into a truth label if the number of timestamps they contain is > the labeling threshold. So a bucket of two raw labels will result in one truth label if there were three users labeling the data file, where the truth label is the earliest of the bucketed records. Thus two raw labels close to each other in time from a single user would count as if two users labeled the one anomaly at slightly different times.

The fix here adds a bucketing step, where the the raw labels from each user are first bucketed such that labels that are w/in the buffer of one another are merged into one raw label (the earliest). Then the rest of the label combing algorithm proceeds.


BoltzmannBrain:
@subutai please review. Note the first anomaly window of ['realRogueAgent/rogue_agent_network_percent.csv'](https://plot.ly/~alavin/2813/raw-data-for-labels-inspection-for-realrogueagentrogue-agent-network-percentcsv/) is removed b/c it overlaps with the probationary period, but changes in a followup PR to the probationary period param should take care of this (i.e. include the anomaly window).

labels/combined_windows.json
```diff
@@ -88,12 +88,7 @@
             "2014-04-15 17:34:00.000000"
         ]
     ],
-    "realAWSCloudwatch/ec2_cpu_utilization_c6585a.csv": [
-        [
-            "2014-04-14 10:34:00.000000",
-            "2014-04-15 20:04:00.000000"
-        ]
-    ],
+    "realAWSCloudwatch/ec2_cpu_utilization_c6585a.csv": [],
```
subutai: What happened to this one?

BoltzmannBrain: It never should've been a truth label, but was included b/c of the bug fixed in commit 49c9aaa. The number of raw label files with an entry for a given data file is used as the number of labelers for that data file. "realAWSCloudwatch/ec2_cpu_utilization_c6585a.csv" is in three of the raw label files, so the algorithm should count three total labelers for this data file. However, only one of the three labelers have an anomaly labeled for this data file, so w/ the old logic the label combining algorithm would count one total labeler for this file. After the bug fix this anomaly label is counted once out of three total labelers, and thus does not pass the threshold.

subutai:
:+1:

subutai:
Merging since @BoltzmannBrain is out sick.

PR #197, opened 2015-09-29
Scoring enhancements

BoltzmannBrain:
@subutai please review. This gives users the option of implementing a flat scoring function rather than the sigmoid. The changes to nab/util.py make sure the results json always reflects the new normalized scores.


subutai:
:+1:

PR #198, opened 2015-09-30
Probationary period maxes out at 750 records

### LINKED ISSUE #114: Probationary period length should not be percentage
Using a percent of data file length is illogical for a streaming data function -- 15% is 600 data points of a file with length 4000, but 3300 for a file length of 2200. Perhaps we set a minimum file size at 1000 (smallest as of v0.8 is 1243, most are ~4000, see issue #113), and then the probationary period is set at 150 data points, although I anticipate larger is called for.

BoltzmannBrain:
@subutai please review. This is the probationary period mod we discussed. I'll email the labelers which files need relabeling.

Fixes #114.


subutai:
:+1:

PR #200, opened 2015-09-30
Normalization cleanup

BoltzmannBrain:


subutai:
Don't we also need to change the checked in results/baseline directory?

BoltzmannBrain:
not yet ready for review

subutai:
:smile:

BoltzmannBrain:
@subutai let's wait until the new data labels are merged in so we can run the null detector and get new results files. Then those results files can be included in this PR.

BoltzmannBrain:
@subutai should be good to go now

subutai:
:+1:

PR #201, opened 2015-09-30
Probationary period calculation in visualizer

BoltzmannBrain:
@subutai please review


BoltzmannBrain:
Merging so we can use this for labeling.

subutai:
:+1:

PR #206, opened 2015-10-08
Improved model params

subutai:
Results:

Final score for 'numenta' detector on 'reward_low_FP_rate' profile = 56.50
Final score for 'numenta' detector on 'reward_low_FN_rate' profile = 69.28
Final score for 'numenta' detector on 'standard' profile = 64.70


subutai:
@BoltzmannBrain

BoltzmannBrain:
:thumbsup:

nab/detectors/numenta/modelParams/model_params.json
```diff
@@ -61,7 +61,7 @@
             "seed": 1956,
             "spVerbosity": 0,
             "spatialImp": "cpp",
-            "synPermActiveInc": 0.002,
+            "synPermActiveInc": 0.003,
```
breznak: @subutai can you wage on the importance (vs. specific-ness) of the changes? I'm just sorting the same problem in https://github.com/breznak/ML.benchmark/issues/14 , and I was assuming that the "state of the art" model params (for generic case) are in `examples/anomaly/temporal/hotgym`. Is it true? Will you propagate this change to nupic upstream?

subutai: @breznak These changes made a meaningful difference to the NAB scores - I believe we should all switch to these parameters for anomaly detection.  There is an API in NuPIC for obtaining the best anomaly detection parameters:

nupic.frameworks.opf.common_models.cluster_params.getScalarMetricWithTimeOfDayAnomalyParams() 

That still needs to be updated with these latest parameters from NAB, see:

https://github.com/numenta/NAB/issues/231

I don't know how good the hotgym parameters are. It should probably use the above API!!

PR #222, opened 2015-10-15
Update scoreboard

BoltzmannBrain:
@subutai 
I'm unsure what to post as the Numenta HTM version.
When we merge this I'll remove the current scoreboard from the wiki.

Edit: Also there's some discussion under the current scoreboard as to why we don't include the Lytics algorithm, but this is a bit lengthy for the readme. What do you suggest we do?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,21 @@ Algorithms – the Numenta Anomaly Benchmark"*, Fourteenth International
 Conference on Machine Learning and Applications, December 2015.
 [[PDF]](http://arxiv.org/abs/1510.03336)
 
+#### Scoreboard
+
+The NAB scores are normalized such that the maximum possible is 100.0 (i.e. the perfect detector), and a baseline of 0.0 is determined by the "null" detector (which makes no detections).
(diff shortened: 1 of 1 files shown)


subutai:
> I'm unsure what to post as the Numenta HTM version.

We know it works with the following SHA's:

NuPIC: 1777c2d27c99be35b49e29a9ddbd81f7b0d082a4
nupic.core: 9b659005cbbd16ee6adc9f2159b130b5ef8a2a82

I think we should put an asterisk and put a footnote saying the results correspond to that version.  The latest version of nupic should still work but may not provide identical results.

For Lytics maybe we can put a short sentence pointing them to a page on the wiki containing the details? In general it is useful and interesting, but the details aren't appropriate for the README.

BoltzmannBrain:
@subutai  take a look now. The Lytics details are now in [this new wiki page](https://github.com/numenta/NAB/wiki/NAB%20Contributions%20Criteria).

subutai:
:+1:

Issue #224, opened 2015-10-22
Allow specifying a subset of the data


chetan51:
Currently if I delete directories in the `data` directory and run NAB, it errors out. I want to be able to run a subset of the data through NAB and get results.

subutai:
I believe you can do that by specifying a different windows file using the command line option `--windowsFile WINDOWSFILE`. It will only score data files mentioned in that windows file.  No need to delete data files. It would be nice to have an example showing that.

chetan51:
@subutai When I try a different windows file with a subset of the entries in it, I get this error when running `run.py`:

  Traceback (most recent call last):
    File "run.py", line 177, in <module>
      main(args)
    File "run.py", line 72, in main
      runner.initialize()
    File "/Users/Chetan/Dropbox/Development/NAB/nab/runner.py", line 93, in initialize
      self.corpusLabel = CorpusLabel(path=self.labelPath, corpus=self.corpus)
    File "/Users/Chetan/Dropbox/Development/NAB/nab/labeler.py", line 119, in __init__
      self.getWindows()
    File "/Users/Chetan/Dropbox/Development/NAB/nab/labeler.py", line 156, in getWindows
      if not all([found(t,data) for t in timestamps]):
    File "/Users/Chetan/Dropbox/Development/NAB/nab/labeler.py", line 132, in found
      f = data["timestamp"][data["timestamp"] == pandas.tslib.Timestamp(t)]
    File "pandas/tslib.pyx", line 298, in pandas.tslib.Timestamp.__new__ (pandas/tslib.c:8967)
    File "pandas/tslib.pyx", line 1146, in pandas.tslib.convert_to_tsobject (pandas/tslib.c:22303)
    File "pandas/tslib.pyx", line 1275, in pandas.tslib.convert_str_to_tsobject (pandas/tslib.c:24301)
  ValueError

subutai:
I ran into something similar before but thought I had fixed it. Apparently not! Is your PR addressing the above issue?

chetan51:
No, my PR is allowing you to delete data files. But if using a different windows file is the correct way to run a subset of the data through, then I will close the PR.

subutai:
Yes, the correct way is to specify a different windows file rather than deleting data files. I think we do want to generate an error if the specified windows file mentions a data file that does not exist.

chetan51:
Okay, I'll try to figure out this error, since it is blocking me on a unicorn task.

chetan51:
[...]

PR #225, opened 2015-10-22
Allow specifying a subset of the data

chetan51:
Fixes https://github.com/numenta/NAB/issues/224.


chetan51:
@subutai Even if you can specify a different windows, I believe this PR adds more flexibility to NAB.

subutai:
> Even if you can specify a different windows, I believe this PR adds more flexibility to NAB.

What precisely are you trying to accomplish? Are you trying to allow the case where the data file does not exist but the windows file has an entry for it?

PR #226, opened 2015-10-27
Fixes #103

### LINKED ISSUE #103: Shouldn't need NuPIC
The readme specifies NuPIC is only needed if running the Numenta detector, yet run.py always runs `from nab.detectors.numenta.numenta_detector import NumentaDetector`, which imports from NuPIC.

SaganBolliger:
@BoltzmannBrain Is this an acceptable solution?


BoltzmannBrain:
:+1:

PR #229, opened 2015-11-30
be more efficient about constructing detector results DF

welch:
Appending rows to DataFrames is expensive. 
This change speeds up the detection phase by around 10X in my testing.

(Even better would have been to pre-allocate np arrays for the DataFrame
constructor, but needed type information isn't available without some plumbing)

### DIFF
--- a/nab/detectors/base.py
+++ b/nab/detectors/base.py
@@ -100,7 +100,7 @@ class AnomalyDetector(object):
 
     headers = self.getHeader()
 
-    ans = pandas.DataFrame(columns=headers)
+    rows = []
     for i, row in self.dataSet.data.iterrows():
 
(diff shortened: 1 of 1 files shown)


welch:
CL signed.

subutai:
@welch Thanks!  This looks like a fine change. We might take a couple of days to approve this - just need to verify the results don't change with this PR.

BoltzmannBrain:
@welch, I've tested the changes and these are good speedups that don't affect the detections and results. Overall the detection phase is sped up, but not 10x because the majority of time is spent within the algorithms code. My testing showed a ~5m speedup for each detector; e.g. the random detector sped up 5m to 13sec, but the skyline detector sped up 285m to 278m.
Good work, thanks!

Issue #235, opened 2016-03-08
NuPIC model params has invalid parameter for RDSE


rhyolight:
The RDSE configuration has a property called [`numBuckets`](https://github.com/numenta/NAB/blob/master/nab/detectors/numenta/modelParams/model_params.json#L45) that is not an input parameter for the actual RDSE. This is confusing. You can see where this value is removed from the actual model params before creating the model and used as part of the resolution calculation [here](https://github.com/numenta/NAB/blob/master/nab/detectors/numenta/numenta_detector.py#L85-L89). 

I found this confusing, as I assumed that the model params could be used as-is.

rhyolight:
I can totally see why this is done, I just think it could be done in a less confusing way. At least make it clear that the `model_params.json` file cannot be uses directly.

BoltzmannBrain:
The RDSE needs to know what granularity at which to setup the data buckets, and then use this granularity to calculate the resolution on a per file basis. That is, `numBuckets` is specified for the entire corpus, from which the RDSE calculates `resolution` for each data file in the corpus. The fix for #231 will make this a bit cleaner.

BoltzmannBrain:
Closed w/ #236

rhyolight:
Thanks @BoltzmannBrain

Issue #234, opened 2016-03-08
Since classifier is not enabled, why have classifier parameters?


rhyolight:
In the [NuPIC model params](https://github.com/numenta/NAB/blob/master/nab/detectors/numenta/modelParams/model_params.json#L21-L27):

```javascript
        "clEnable": false,
        "clParams": {
            "alpha": 0.035828933612158,
            "clVerbosity": 0,
            "regionName": "CLAClassifierRegion",
            "steps": "1"
        },
```

Since `clEnable` is false, can't you just remove the whole `clParams` block?

BoltzmannBrain:
Yes, the `clParams` dict can be removed, but the `clEnable` entry must remain as it is expected by the CLA model. Including this there are a few issues around nab/nupic params (#231, #235), so I'll wrap them together into a future PR.

BoltzmannBrain:
Closed w/ #236

PR #236, opened 2016-03-10
HTM detector use NuPIC params

### LINKED ISSUE #231: NAB should be updating and using nupic's getScalarMetricWithTimeOfDayAnomalyParams
NAB PR #206 recently made improvements to its own copy of the model parameters.

Htmengine is relying on  nupic.frameworks.opf.common_models.cluster_params.getScalarMetricWithTimeOfDayAnomalyParams() for its model parameters.  

NAB should be using and updating nupic.frameworks.opf.common_models.cluster_params.getScalarMetricWithTimeOfDayAnomalyParams instead of maintaining its own copy of the parameters. Similar idea to "Switch over to using the Anomaly Likelihood class in NuPIC” (https://github.com/numenta/NAB/pull/184).

Per email from Subutai:

> Subject: Re: HTM Engine questions on mailing list
>
> Vitaly,
>
>Yes, I think that would be better.  We should update the params in common_models and NAB should switch over to using it.
>
> -- Subutai

BoltzmannBrain:
Fixes #231, #234, #235


BoltzmannBrain:
@subutai please review. I'll followup with new results and scores.

nab/detectors/numenta/numenta_detector.py
```diff
@@ -66,28 +68,12 @@ def handleRecord(self, inputData):
 
 
   def initialize(self):
-    calcRange = abs(self.inputMax - self.inputMin)
-    calcPad = calcRange * .2
+    # Get config params, setting the RDSE resolution
```
subutai: I'm not sure this strategy matches the NAB requirements. Currently you are not allowed to look ahead and use data values to set your parameters. The only exception is that we assume min/max for each metric are known to the detector (this is a reasonable assumption in most practical scenarios).  Passing in the entire dataset to some black box to set parameters however is not allowed.

nab/detectors/numenta/numenta_detector.py
```diff
@@ -66,28 +68,18 @@ def handleRecord(self, inputData):
...
+    rangePadding = abs(self.inputMax - self.inputMin) * 0.2
```
BoltzmannBrain: @subutai I tried several different values for the range padding and found 0.2 to give the best results.

BoltzmannBrain:
@subutai scores increased to 65.27, 58.59, and 69.38 (standard, low FP, low FN). Good to merge?

subutai:
:+1: OK to merge (after the nupic merge is complete).  Thanks!

PR #237, opened 2016-04-02
Add 'htm.java' detector

lscheinkman:
Add **htm.java** detector to NAB by modifying **NumentaDetector** to use **htm.java** model instead of **OPF** model to get the raw anomaly values. All other logic remains the same as the original **NumentaDetector**. In the future we should also replace the **anomaly likelihood** logic to use the **htm.java** implementation.

Follow the `README` instructions to build and run this new detector.

In this initial implementation I've mapped **OPF** model parameters directly to **htm.java** parameters, however we may need to fine tune **htm.java** parameters to get better results.

@cogmission Could you please review and if possible help me find out what is best set of parameters? I think I am missing some basic settings because the results I am getting are very different from the `python` version.

@BoltzmannBrain Please review as well.

CC: @subutai , @rhyolight


nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
```diff
@@ -0,0 +1,266 @@
...
+                            .add(Anomaly.create())
+                            .add(new TemporalMemory()))
+                    .add(Network.createLayer("SpatialPooler", parameters)
+                            .add(new SpatialPooler()))
+                    .add(Network.createLayer("Sensor", parameters)
+                            .add(Sensor.create(ObservableSensor::create,
+                                    SensorParams.create(SensorParams.Keys::obs, "Manual Input", publisher))))
+                    .connect("TemporalMemory", "SpatialPooler")
+                    .connect("SpatialPooler", "Sensor"));
+
```
cogmission: While adding the algorithms to their own "Layers" is totally legitimate, I wouldn't do it in production unless it served a particular purpose because it will run a bit slower. An alternative would be: (no need to connect anything here).

```Java
network = Network.create("NAB Network", parameters)
                 .add(Network.createRegion("NAB Region")
                    .add(Network.createLayer("SheinkmanLayer", parameters)
                             .add(Anomaly.create())
                             .add(new TemporalMemory())
                             .add(new SpatialPooler())
                             .add(Sensor.create(ObservableSensor::create,
                                     SensorParams.create(SensorParams.Keys::obs, "Manual Input", publisher))));

**Comment by cogmission** [comment]
@lscheinkman This is AWESOME! :+1: 

Nice to meet you!

I left a line note here: https://github.com/numenta/NAB/pull/237/files#diff-d53aa77236428fba395772cc18304a64R59

In general the code looks very good... I'm curious about piping your input from standard in, and whether you will be entering the data from the command line? So depending on your source, there is also the FileSensor which can read a file in directly (obviating the need for a Publisher)? As I said, I'm not sure what your intentions are and also I'm not at all familiar with NAB. 

I'm sure we can bounce ideas back and forth and arrive at the best treatment... But overall I'm very impressed with the use of the API - What were your experiences getting in to it? Did you find it difficult to figure things out? How was the documentation? Any feedback you can give me to improve things would be greatly appreciated!

Regarding the Parameters. The parameters used in HTM.java are with few exceptions directly taken from the NuPIC algorithms (i.e. no NetworkAPI or OPF variables were used). So as far as performance or output goes, the engineers at Numenta should be able to help you come up with the best parameters to elicit the best overall output.

Let me know how else I can help?

**Comment by lscheinkman** [comment]
@cogmission We've met before when you came to our office during the "HTM Challenge" last year. I'm the Numenta engineer who was sitting next to you in the office. 

Great job on the Network API!
The API is simple, intuitive and really easy to use. I was able to get a model up and running really quick just by looking at some of the examples. My overall experience with `htm.java` was great. I was very impressed on how quick and easy I was able to create a full HTM Java project from zero.

Thank you for the quick feedback, I will simplify the network by removing the extra layers. I was just trying to mimic the OPF network, but now I understand `htm.java` does not need the extra regions and layers for this simple network.

As far as piping the `stdin/stdout`, it was done to simplify the communication with the NAB framework. This class was written so it could be called by NAB from a `python` process. See [htmjava_detector.py](https://github.com/lscheinkman/NAB/blob/htm.java/nab/detectors/htmjava/htmjava_detector.py) for more details.

Regarding the parameters, I am using the model parameters extracted from **nupic** via [getScalarMetricWithTimeOfDayAnomalyParams](https://github.com/numenta/nupic/blob/master/src/nupic/frameworks/opf/common_models/cluster_params.py#L29). Based on what you said, my guess is that OPF must be changing the default values for other algorithm parameters not returned by this function. I will double check OPF parameters and updated any missing algorithm parameter value with OPF default value and let you know if that improves the results.

Thank you for your help.

**Comment by cogmission** [comment]
@lscheinkman Thanks for the feedback! I just thought of another potential deviation between versions, which could be if the NAB's configuration includes the old TP (htm.java doesn't use the old TP but rather the new TemporalMemory)?

File: README.md [hunk 2]
```diff
@@ -158,6 +158,21 @@ the specific version of NuPIC (and associated nupic.core) that is noted in the
...
+##### Run htm.java with NAB
+
+First make sure you have __java 8__ installed.
+ 
+    java -version
+        
+Build __htm.java__ NAB detector:
+  
+    cd nab/detectors/htmjava
+    ./gradlew clean build
```
BoltzmannBrain: Does this need to be run every time, or just the first time someone wants to use the htmjava detector?

nab/detectors/htmjava/htmjava_detector.py
```diff
@@ -0,0 +1,130 @@
+# ----------------------------------------------------------------------
+# Copyright (C) 2014, Numenta, Inc.  Unless you have an agreement
```
BoltzmannBrain: 2016

nab/detectors/htmjava/htmjava_detector.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # Initialize the anomaly likelihood object
+    numentaLearningPeriod = math.floor(self.probationaryPeriod / 2.0)
+    self.anomalyLikelihood = anomaly_likelihood.AnomalyLikelihood(
+      claLearningPeriod=numentaLearningPeriod,
+      estimationSamples=self.probationaryPeriod-numentaLearningPeriod,
+      reestimationPeriod=100
+    )
+
+    if self.model:
```
BoltzmannBrain: What is this for? I'm assuming to close the pipe.
It would be better to have a `_self.terminateJavaPipe()`method that does this, and is also called at the end of `run()`.

nab/detectors/htmjava/htmjava_detector.py
```diff
@@ -0,0 +1,130 @@
...
+    encoderParams["timestamp_dayOfWeek"] = encoderParams.pop("c0_dayOfWeek")
+    encoderParams["timestamp_timeOfDay"] = encoderParams.pop("c0_timeOfDay")
+    encoderParams["timestamp_timeOfDay"]["fieldname"] = "timestamp"
+    encoderParams["timestamp_timeOfDay"]["name"] = "timestamp"
+    encoderParams["timestamp_weekend"] = encoderParams.pop("c0_weekend")
+    encoderParams["value"] = encoderParams.pop("c1")
+    encoderParams["value"]["fieldname"] = "value"
+    encoderParams["value"]["name"] = "value"
+
+    self.sensorParams = encoderParams["value"]
```
BoltzmannBrain: Please add newline at end of files.

BoltzmannBrain:
@lscheinkman I left a few comments, but overall looks good.
We still will hold off merging until we can validate htmjava yields the same results as the NuPIC detector.

cogmission:
@BoltzmannBrain You'll never get the _**exact**_ same result as we don't share RNG implementations... and the Mersenne Algorithm while purported to be the same, differs greatly - and then you guys don't call the language neutral methods on the RNG to top it off! ;-P

I got frustrated...

Then when I discovered how god awful slow the algorithm is, I stopped using git altogether. You guys should check out HTM.java's fast RNG implementation...  [(see here)](https://github.com/numenta/htm.java/blob/master/src/main/java/org/numenta/nupic/util/FastRandom.java)

subutai:
> You'll never get the exact same result as we don't share RNG implementations

That's fine. Right now the NAB score of the implementation using HTM.Java is terrible - it is basically random. We just need to debug, figure out the issues, and get the score in the same ballpark.  It's probably something simple - I'm confident @lscheinkman or @cogmission will figure it out!

cogmission:
You're never going to get the _**same**_ results in HTM.java as we don't share RNG implementations. While the Mersenne algorithm purports to be the same, one has to use the language neutral methods (the general nextInt() method) or else we can't get the same result. You guys use the more "fancy" methods and so there is no compatibility to be had.

Also, there is a faster algorithm which is easily portable to Python or C++ [here...](https://github.com/numenta/htm.java/blob/master/src/main/java/org/numenta/nupic/util/FastRandom.java#L71) - you guys might consider using it?

subutai:
> You're never going to get the same results in HTM.java

I agree with you (please read my comment). However we do want to get decent results in HTM.java.

cogmission:
Awesome! :-) I'll never turn down care and lovin' :-P

cogmission:
@lscheinkman any news on what's going on with HTM.java and NAB yet? No rush... I'm just anxious to see what the problem is? :-)

lscheinkman:
@cogmission Sorry for going slow on this PR, got busy on other tasks. I will try to debug this week and keep you posted. 
On thing I've noticed was using `KEY.MAX_BOOST` parameter default value (10.0) instead of NAB's model value (1.0) improves performance but still not comparable to `TM_CPP` yet.

cogmission:
@lscheinkman Thanks! Please keep me posted (I'm rooting for ya') :+1:

lscheinkman:
Once https://github.com/numenta/nupic/pull/3138 is merged we should try the new TM parameters and see if we get similar results

cogmission:
@lscheinkman +1

BoltzmannBrain:
@lscheinkman @cogmission would one of you be able to try the new TM params?

lscheinkman:
@BoltzmannBrain I will

lscheinkman:
@cogmission @BoltzmannBrain @subutai 
Here are the results with the new parameters:
```
    "htmjava": {
        "reward_low_FN_rate": 12.80037041235632,
        "reward_low_FP_rate": 3.7401232426724107,
        "standard": 9.28676251491379
    }
```

cogmission:
@lscheinkman What do these mean? :-P

BoltzmannBrain:
@cogmission these are to be compared with the results of the new TM detector (#240):

```
    "numentaTM": {
        "reward_low_FN_rate": 66.09146564859769,
        "reward_low_FP_rate": 52.389446324344824,
        "standard": 61.20616399012932
    },
```

We would expect the HTMJava detector to match these scores.

cogmission:
@lscheinkman So then, what's the conclusion? Do new TM parameters correspond to new TM code? or are they just new parameters? If they correspond to new code, then HTM.java doesn't have it... Otherwise, I don't know what to say - if HTM.java is unit test and integration test identical - then what gives? I wonder...?

lscheinkman:
@cogmission AFAIK there is no new TM code,  numenta/nupic#3138 only updates TM parameters.
Unit and integration tests validates the code implementation not the algorithm performance. 

To test algorithm performance you need to compare the results from different implementation, and that's what `NAB` is doing.

One way to test the algorithm is to compare `hotgym` results and see if they are similar. 
Currently `nupic` runs  [benchmarks tests](https://github.com/numenta/nupic/blob/master/tests/regression/run_opf_benchmarks_test.py) where it validates the prediction results as part of its regression tests.
Another example can be found in [HTM Studio](https://github.com/numenta/numenta-apps/tree/master/unicorn/tests/py/integration/compatibility_test) where the test uses `NAB` results as ground truth.

CC: @subutai

cogmission:
> Unit and integration tests validates the code implementation not the algorithm performance.
> 
It seems to me that the same algorithm written in different languages, if 100% determinant, should perform identically? ;-)

lscheinkman:
Assuming 100% test coverage and no differences between the implementations.

Basically this difference in the results indicates a bug somewhere not covered by the tests, either on my use of `HTM Java` in `NAB` or some implementation differences between `java` and `python` that would only manifest in certain situations.

We can start to investigating by comparing `hotgym` results using both `python` and `java` implementation and later on move to other `NAB` results.

cogmission:
Sounds good. Does NAB use Anomaly Likelihood or just the Anomaly call? Also we can piece together the algorithms absent of any NAPI implementation and do metrics on them in raw form with just the Encoder, SP, TM, and Anomaly classes. 

I took great care to ensure 100% test viability between _existing_ tests, but I can't account for NuPIC itself not having 100% test coverage. Not that I'm expecting 100% test coverage or criticizing.

Anyway, that would be my recommendation.

fergalbyrne:
@lscheinkman I believe the problem is in the HTM.java code. Currently believe it has to do with the construction and mutation of ManualInput observer objects from the TM output in the Layer code. Suspicion is that previousPredictedCells is being copied twice, overwritten or skipped by the TemporalFunc in Layer. In other words, the whole Anomaly infrastructure on HTM.java is a bit broken. Chatting with @cogmission about this and we'll see if we can track it down ASAP.

cogmission:
> In other words, the whole Anomaly infrastructure on HTM.java is a bit broken. 

This is incorrect and over stated. 

I could have done without the dramatic summary. One line of code is suspect. There are actually 2 I believe in symmetrical places (i.e. the same return value in the SpatialPoolerFunc also that should be "unravelled" also)

fergalbyrne:
@cogmission let's just park the histrionics. @BoltzmannBrain (the expert on NAB) says that HTM.java is producing completely wrong scores, when it should be responding like NuPIC does, so let's find the source of that. Is there a reason not to merge this PR (once conflicts are resolved) so we can run HTM.java on NAB and check for ourselves?

cogmission:
@fergalbyrne Correction, @BoltzmannBrain is saying the NAB Detector for HTM.Java is producing wrong scores. It could be attributed to HTM.Java proper, or not.

You, my friend are jumping to conclusions and slinging speculation left and right. I don't deny that it _could_ be the NAPI, but it definitely doesn't warrant your indiscriminate criticisms when you have no more clue than anyone else, what the cause is. Enough said... I agree lets get to work.

BoltzmannBrain:
This PR needs reviewing before it can be merged. @fergalbyrne or @cogmission would you be able to look through the use of HTM.java here? I don't now java well enough to give a thorough review. We need to be confident this detector is a correct use of HTM.java in order to assert that HTM.java yields these NAB scores.

@fergalbyrne if you want to run this detector before merging just fetch this PR.

cogmission:
@BoltzmannBrain Cortical gave me the green light to spend some time on this, so yes I'll review this PR and also I'll determine whether there is something amiss in the NAPI where Anomaly functionality is concerned. NAB coding from that point will have to be a hand-off to you or @lscheinkman from there (I'll guide as much as possible).

Also I'll open an issue to clean up the 2 lines of code @fergalbyrne brought to my attention and vet them to make sure the squirrelliness isn't just stylistic. I ran this code with extreme rigor against a non-network API assembly to make sure I got same results long ago (when I was vetting the mixin framework and new TM implementation for Yuwei's orphan segment code; which is why I refuse to rollover for suspicions based on "smell" alone that aren't corroborated by proof) - but as @rhyolight says, everyone encounters bugs - I just refuse to abide the extraneous conclusions and inaccurate language when we don't know for sure. Once it's proven, then peeps can cast aspersions all they like, if that's their thing.

rhyolight:
> I just refuse to abide the extraneous conclusions and inaccurate language when we don't know for sure. Once it's proven, then peeps can cast aspersions all they like, if that's their thing.

@cogmission David, my friend, take a break and enjoy some birthday cake. 🍰  🎂  🎉

BoltzmannBrain:
@cogmission as another option, if you would prefer creating a new PR with an HTM.java detector that you implement, rather than vetting this PR, please go ahead 😀

cogmission:
@BoltzmannBrain I won't be able to take a look at this until tomorrow. For now, I really can't say and my instinct is to start with the work that is already done?

BoltzmannBrain:
Up to you @cogmission.

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
```diff
@@ -0,0 +1,383 @@
...
+    public HTMModel(JsonNode modelParams) {
+        LOGGER.trace("HTMModel({})", modelParams);
+
+        // Create Sensor publisher to push NAB input data to network
+        publisher = Publisher.builder()
+                .addHeader("timestamp,value")
+                .addHeader("datetime,float")
+                .addHeader("T,B")
+                .build();
+
```
cogmission: Hi Luiz,

As part of the new Persistence API, the idiom for creating the ```Publisher``` has changed. The "new" (or more advanced) way to add a ```PublisherSupplier``` instead.

So you can substitute this code for lines: 45 - 49...

``` Java
PublisherSupplier supplier = PublisherSupplier.builder()
            .addHeader("timestamp,value")
            .addHeader("datetime,float")
            .addHeader("T,B")
            .build();
```
...and then on line 62, add the ```supplier``` instead of the ```publisher```.

That should do it! ;-)

BoltzmannBrain:
Just an update, I ran this (with the most recent nupic):

```
Final score for 'htmjava' detector on 'reward_low_FP_rate' profile = 3.46
Final score for 'htmjava' detector on 'reward_low_FN_rate' profile = 13.00
Final score for 'htmjava' detector on 'standard' profile = 9.59
```

cogmission:
@BoltzmannBrain Thank you for running that! :-)

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
```diff
@@ -138,9 +144,13 @@ public Parameters getSpatialPoolerParams(JsonNode modelParams) {
...
-        if (spParams.has("seed")) {
-            p.set(KEY.SEED, spParams.get("seed").asInt());
-        }
+        // if (spParams.has("seed")) {
+        //     p.set(KEY.SEED, spParams.get("seed").asInt());
+        // }
+        // For some seeds, htm.java throws
+        // java.lang.ArrayIndexOutOfBoundsException: 1017
+        //    at org.numenta.nupic.Connections.computeActivity(Connections.java:1199)
+        p.set(KEY.SEED, 42);
```
cogmission: I think it would be better to open an Issue in HTM.Java, than making this comment a permanent feature inside this class. I don't think there's any reason to document bugs. They should just be fixed. IMHO

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
```diff
@@ -194,9 +204,12 @@ public Parameters getTemporalMemoryParams(JsonNode modelParams) {
...
         }
-        if (tpParams.has("seed")) {
-            p.set(KEY.SEED, tpParams.get("seed").asInt());
-        }
+        // if (tpParams.has("seed")) {
+        //     p.set(KEY.SEED, tpParams.get("seed").asInt());
+        // }
+        // For some seeds, htm.java throws
+        // java.lang.ArrayIndexOutOfBoundsException: 1017
+        //    at org.numenta.nupic.Connections.computeActivity(Connections.java:1199)
```
cogmission: I think it would be better to open an Issue in HTM.Java, than making this comment a permanent feature inside this class. I don't think there's any reason to document bugs. They should just be fixed. IMHO

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
```diff
@@ -60,15 +61,23 @@ public HTMModel(JsonNode modelParams) {
         // Get updated model parameters
         Parameters parameters = getModelParameters(modelParams);
 
+        HTMSensor<Object> sensor = (HTMSensor<Object>)Sensor.create(ObservableSensor::create,
+                SensorParams.create(SensorParams.Keys::obs, "ManualInput", supplier));
+
+        sensor.initEncoder(parameters);
+
+        int inputWidth = sensor.getEncoder().getN();
+        parameters.set(KEY.INPUT_DIMENSIONS, new int[]{inputWidth});
```
cogmission: This key is moot. The NAPI will set this automatically

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
mrcslws: Sure, I can change it to a TODO comment linking to the github issue.

Yeah, it's weird that we're passing in parameters that have multiple seeds while htm.java only uses one seed. The same thing happens with the column dimensions. I think a future PR could do some cleanup here.

cogmission:
@mrcslws 
Also Marcus, thank you so much for identifying that silent derived configuration, I would never have found that! I really appreciate you coming in and realizing that. You should have a cape or something! ;-)

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
mrcslws: Ouch. I put a lot of time into figuring out how to get that number.

mrcslws: Oh, I needed the number anyway for the potential radius... I think I might leave it as-is since it makes it clear that we're matching the potential radius to the input dimensions. And then a future change could maybe change the handling of the potential radius. (For example, maybe a "potentialRadius" of -1 should mean "the whole input space" and then all this pain and suffering can go away)

cogmission:
@lscheinkman @mrcslws 

I got these scores:
```
Running score normalization step
baselines[reward_low_FP_rate] = -116.0
baselines[reward_low_FN_rate] = -232.0
baselines[standard] = -116.0
Final score for 'htmjava' detector on 'reward_low_FP_rate' profile = 49.97
Final score for 'htmjava' detector on 'reward_low_FN_rate' profile = 63.08
Final score for 'htmjava' detector on 'standard' profile = 58.41
Final scores have been written to /Users/cogmission/git/lscheinkman/NAB/results/final_results.json.
```

Which are a little better than Marcus'. and in the final results file I got:
```
 "htmjava": {
        "reward_low_FN_rate": 63.07649796008621,
        "reward_low_FP_rate": 49.971153542524476,
        "standard": 58.40785038840517
    },
```

So obviously there's some room to optimize the settings? (I ran optimize, but I'm just thinking about all settings).

I would like to show HTM.Java in the best light possible for any posted scores on the "official" table in NAB's README.md file - before posting them ok?

nab/detectors/htmjava/src/main/java/nab/detectors/htmjava/HTMModel.java
cogmission: > Ouch.

Don't feel bad. Your code showed great insight into the HTM.Java's Network API, I was impressed seeing the HTMSensor reference 😉 

> Oh, I needed the number anyway for the potential radius... I think I might leave it as-is since it makes it clear that we're matching the potential radius to the input dimensions. And then a future change could maybe change the handling of the potential radius. (For example, maybe a "potentialRadius" of -1 should mean "the whole input space" and then all this pain and suffering can go away)

I think you're right. It should be set internally and "-1" as an option does sound like a good idea!

mrcslws:
@cogmission How about we merge this PR in its current state (assuming others approve), then you can submit another PR that adds it to the README and does any other cleanup that you'd like to see happen?

README.md
```diff
@@ -35,11 +35,12 @@ The NAB scores are normalized such that the maximum possible is 100.0 (i.e. the
 |---------------|------------------|---------------|---------------|
 | Perfect       | 100.0            | 100.0         | 100.0         |
 | [Numenta HTM](https://github.com/numenta/nupic)* | 65.3          | 58.6          | 69.4          |
+| [Relative Entropy] (http://www.hpl.hp.com/techreports/2011/HPL-2011-8.pdf) | 54.6 | 47.6 | 58.8 |
 | [Twitter ADVec v1.0.0](https://github.com/twitter/AnomalyDetection)| 47.1             | 33.6          | 53.5          |
 | [Etsy Skyline](https://github.com/etsy/skyline) | 35.7             | 27.1          | 44.5          |
+| [Sliding Threshold](https://github.com/numenta/NAB/blob/master/nab/detectors/gaussian/windowedGaussian_detector.py) | 30.7             | 12.1         | 38.3          |
```
mrcslws: @lscheinkman before merging, it might be worth double-checking this isn't clobbering any values in the README.

BoltzmannBrain:
I think the readme should be revised:

- The htm.java results should go in the "add'l flavors of HTM" table (along with an embedded link to the htm.java repo), not the main scoreboard
- The instructions on how to run the htm.java detector on NAB should go in a readme in the detectors/htmjava dir. There we could also clarify an extra dependencies needed to run the htm.java detector

rhyolight:
Great work you guys!!! I am so looking forward to putting this baby to bed.

lscheinkman:
@subutai I've updated the README with the results. Let me know if it is ok to merge.

mrcslws:
@lscheinkman I don't think we want to add it to the README in this PR. See comment above. https://github.com/numenta/NAB/pull/237#issuecomment-250072897

lscheinkman:
@mrcslws I was following the feedback from https://github.com/numenta/NAB/pull/237#issuecomment-250169573

README.md
```diff
@@ -85,13 +85,15 @@ run without likelihood, set the variable `self.useLikelihood` in
...
 * [nab-comportex](https://github.com/floybix/nab-comportex) is a twist on HTM  anomaly detection using [Comportex](https://github.com/htm-community/comportex), a community-driven HTM implementation in Clojure. Please see [Felix Andrew's blog post](http://floybix.github.io/2016/07/01/attempting-nab) on experiments with this algorithm.
+* [HTM Java](https://github.com/numenta/htm.java) is a Community-Driven Java port of HTM.
 
 
 | Detector      |Standard Profile | Reward Low FP | Reward Low FN |
 |---------------|---------|------------------|---------------|---------------|
 | Numenta HTM*   | 65.3             | 58.6       | 69.4          |
 | [nab-comportex](https://github.com/floybix/nab-comportex)&dagger; | 64.6             | 58.8       | 69.6          |
 | [NumentaTM HTM](https://github.com/numenta/NAB/blob/master/nab/detectors/numenta/numentaTM_detector.py)* | 61.2             | 52.4       | 66.1          |
+| [HTM Java](https://github.com/numenta/NAB/blob/master/nab/detectors/htmjava) | 57.54 | 49.71 | 62.33 |
```
cogmission: Guys,

Can we please get HTM.Java's results = NuPIC before we post? Please take this down? There's no reason why HTM.Java's results have to be less than NuPIC, so why would we post these without optimizing it?

lscheinkman: @cogmission Feel free to push your `HTMModel` changes and results to this PR. You should have access to it.
I will not merge until you are satisfied with the results. 
BTW, **HTM Java** implementation should be compared with **NumentaTM HTM** results which are pretty close right now.

lscheinkman: Done. I've unpublished the current HTM Java results from the PR.
IMHO, we should not merge to master until we get results we can publish.

cogmission: So who is going to own the work on the optimization? That is Numenta's "wheelhouse" not mine?

subutai:
@lscheinkman I haven't followed all the ins and outs of this PR, but I agree that the results should be roughly on par with "NumentaTM HTM". We also shouldn't merge this PR until we have results we are happy with in the leaderboard. (i.e. I don't want the algorithm in the NAB repo without publishing results.)  I don't mind waiting until it has been tuned more.

cogmission:
@lscheinkman 
I will help as much as you like - for instance I will do a release today containing the fix within HTM.Java that internally configures the ```potentialRadius``` parameter that was throwing things off. Then you can build to a release version v0.6.9-alpha. 

After that, I can do a diff of my local HTMModel.java file (which should be the same or nearly as the one @mrcslws submitted since he used my file as a baseline) - and push it to this PR.

I will announce here when the above is completed so that you can begin looking at parameter tweaks.

cogmission:
@lscheinkman (or optimization effort owner whoever that is?)

Currently I'm working on merging my ```HTMModel``` differences and running it to see why mine has higher scores. Then I'll push to this PR so we have something to "share" which you or someone else can optimize.

mrcslws:
These scores are good. I've posted the latest numentaTM scores here: https://github.com/numenta/NAB/issues/276

cogmission:
Do we have any insight into what contributed to the drop in scores since May (nupic.core SHA c030b84)?

cogmission:
@mrcslws @subutai 
These scores look compatible with [#276](https://github.com/numenta/NAB/issues/276). Not sure why the drop from May, but my only interest is maintaining compatibility with NuPIC. So I'm good with these unless you are going to optimize NuPIC, in which case I should do the same for HTM.Java...

lscheinkman:
@cogmission  Please review the README with the published results. Let me know if you would like to change anything before we merge

cogmission:
@lscheinkman I'll defer to @subutai because it means NuPIC's scores have to be lowered too, no?

mrcslws:
The "numenta" scores have also fallen. I'm investigating which commit caused it to drop. I think it was one of my recent changes to the SP.

cogmission:
@mrcslws @lscheinkman I pushed changes to the build; which integrates the latest HTM.Java release. The new release internally sets the ```potentialRadius``` variable within the Network aPI, making the local changes to ```HTMModel``` which also set that variable, unnecessary.

lscheinkman:
@cogmission The merge commit ddf61ee removed all NAB results from the repo. 
Could you revert that one and submit only just changes related to `nab/detectors/htmjava` directory?

cogmission:
@lscheinkman @Everyone :-)

Aren't we going to merge this after the long bout in hell we spent getting this to work?

lscheinkman:
@cogmission LGTM, I believe we are just waiting on #276. There is an open PR fixing it already. See #277 

Let me know If you are ok with the current results in the README so we could merge this PR now and not wait for #276.

CC: @subutai

cogmission:
The scores should match very closely with NuPIC's. Can you re-run it using 0.6.10 (build.gradle version) and see if they match. They should be within 0.2 points I would say?

lscheinkman:
@cogmission The current scores already match the latest **NumentaTM** scores. See #276 for updated values.
The current values for **NumentaTM** in the [README](https://github.com/numenta/NAB/blob/master/README.md#additional-scores) reflects the scores taken from nupic (numenta/nupic@42f701d) and nupic.core (numenta/nupic.core@c030b84) versions. PR #277 will update these values to the latest version.

I will leave up to you whether we should merge this PR before #277

cogmission:
@lscheinkman 

>  I will leave up to you whether we should merge this PR before #277  

Let's wait for #277 so that there is no unexplained difference in the scores, ok?

cogmission:
@rhyolight @mrcslws @lscheinkman Just a friendly reminder that Marcus [was going to look into](https://github.com/numenta/NAB/pull/237#issuecomment-250550883) why the scores dropped (possibly due to a change in the SP?) - I don't think we should lose track of this task as its important to figure out... Not sure if this "reminder" is warranted or not because Marcus may have already looked into this?

rhyolight:
t

mrcslws:
@cogmission Here was the result of that investigation: https://github.com/numenta/NAB/issues/245#issuecomment-250803245

The solution is probably to post the mean score across different random seeds.

scottpurdy:
See #287 for updated numentaTM results. There is a spatial detection change to the inherited NumentaDetector class that will need to be done in order to get similar results:

https://github.com/numenta/NAB/blob/cf1b9eb7d00a5714a666b3c3890102a034344a06/nab/detectors/numenta/numenta_detector.py#L103-L104

lscheinkman:
@cogmission I've updated the detector with  @scottpurdy changes and now **htm.java** score is:
```
    "htmjava": {
        "reward_low_FN_rate": 70.42407766520115,
        "reward_low_FP_rate": 53.25694971610345,
        "standard": 65.54990960125
    }
```
I've updated the [README.md](https://github.com/lscheinkman/NAB/blob/08fc1aad8ae55565d270c262323bae0b1d2df9d7/README.md) with the latest values.

Please review and let me know if it is Ok to merge.

subutai:
@lscheinkman @scottpurdy Nice work!

cogmission:
@lscheinkman @scottpurdy Indeed, very nice! I do believe a merge is in order!

cogmission:
@lscheinkman Sorry for the late comment, but shouldn't the code beginning at [line 95](https://github.com/numenta/NAB/pull/237/commits/8b5f0d4bd906eede46bf971b159b009626940d12#diff-121107e04b7afd3a97107a5af00b460aR95) be _**above**_ the call to the anomaly likelihood code at [line 92](https://github.com/numenta/NAB/pull/237/commits/8b5f0d4bd906eede46bf971b159b009626940d12#diff-121107e04b7afd3a97107a5af00b460aR92) to avoid executing the likelihood code for efficiency (since the ```logScore = 1.0``` value isn't affected by the call to the likelihood code?

lscheinkman:
@cogmission Good catch!
I will add new PR fixing this issue here and the [numenta_detector.py](https://github.com/numenta/NAB/blob/master/nab/detectors/numenta/numenta_detector.py#L103)

lscheinkman:
@cogmission  See https://github.com/numenta/NAB/pull/288#discussion_r90074319

PR #242, opened 2016-06-03
RES-210 Add sliding threshold baseline detector

zuhaagha:
Implementation of a sliding threshold baseline detector that calculates a gaussian distribution over a window of previous data points. For now, the anomaly score of the detector is the tail probability of the gaussian based on Q-function using the method from nupic's anomaly_likelihood.py but it is subject to change based on feedback. 

I have tried some combinations of windowSize and stepSize but the parameters are not fully tuned yet, so the final scores of the detector may not be optimal. I am creating a separate jira subtask for tuning.


zuhaagha:
@subutai @BoltzmannBrain Please review.

zuhaagha:
NAB score for this detector currently less than the random detector

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,73 @@
...
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import numpy as np
+from nab.detectors.base import AnomalyDetector
+from nupic.algorithms import anomaly_likelihood
+
+
+class WindowedGaussianDetector(AnomalyDetector):
```
subutai: Class comment describing the algorithm.

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def handleRecord(self, inputData):
+        """Returns a tuple (anomalyScore).
+        The anomalyScore is the tail probability of the gaussian (normal) distribution over a sliding window of inputData values. The tail probability is based on the Q-function.
+        """
+
+        anomalyScore = 0.0
+        inputValue = inputData["value"]
+        if len(self.windowData) > 0:
+            # anomalyScore = np.abs(inputValue - self.mean) / self.std
```
subutai: Shouldn't have commented out code.

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def handleRecord(self, inputData):
+        """Returns a tuple (anomalyScore).
+        The anomalyScore is the tail probability of the gaussian (normal) distribution over a sliding window of inputData values. The tail probability is based on the Q-function.
+        """
+
+        anomalyScore = 0.0
+        inputValue = inputData["value"]
+        if len(self.windowData) > 0:
+            anomalyScore = anomaly_likelihood.normalProbability(inputValue, {"mean": self.mean, "stdev": self.std})
```
subutai: Nice reuse of existing functionality!

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,76 @@
...
+
+        windowUpdated = False
```
subutai: Use of a flag windowUpdated is hacky and makes the logic harder to follow. Why not just create a private method _updateWindow that contains lines 68-71 and call it wherever you set `windowUpdated=True`

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,76 @@
...
+                self.windowData = self.windowData[self.stepSize:]
+                self.windowData.extend(self.stepBuffer)
+                windowUpdated = True
+                # reset stepBuffer
+                self.stepBuffer = []
+
+        if windowUpdated:
+            self.mean = np.mean(self.windowData)
+            self.std = np.std(self.windowData)
+            if self.std == 0.0:
```
subutai: I'm concerned that this is not the right handling of this case. Suppose you have percent CPU usage, which goes from 0.0 to 1.0. If the CPU is flat for 20 iterations, the std returned by numpy will be 0 and self.std will be set to 1. If the CPU changes, that should be considered an anomaly, but it won't because the std is too large for that range. 

I would recommend setting it to a very tiny value, such as 0.000001

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+        A sliding window detector that computes anomaly score of a data point by computing its probability from the gaussian distribution over a window of previous data points.
+    """
+
+    def __init__(self, *args, **kwargs):
+
+        # Initialize the parent
+        super(WindowedGaussianDetector, self).__init__(*args, **kwargs)
+
+        self.windowSize = 20
```
subutai: Put comment that windowSize was tuned to give best performance on NAB.

subutai:
@zuhaagha Have you visually looked at some of the results and does the detector behave as you would expect it to?  There are some anomalies, such as obvious spikes, which this detector should get.

zuhaagha:
> Have you visually looked at some of the results and does the detector behave as you would expect it to?

That's a good idea! I am going to try that.

nab/detectors/gaussian/windowedGaussian_detector.py
zuhaagha: Yes, I was unsure about it as well. But I was thinking let's say I have [0.1,0.1,0.1...] for 20 iterations and then I receive a [1] which should be treated as an anomaly. The mean is 0 and std is 0.000001. So the 1/0.000001 will be a very large score but the Q-function returns 0 for any score greater than 70, so it won't be treated as an anomaly. But let me try how it does and plot the graphs.

results/stdout.txt
```diff
@@ -0,0 +1,219 @@
+
```
BoltzmannBrain: Why is this committed?

zuhaagha: Oops! That's a mistake.

BoltzmannBrain:
@zuhaagha a couple general comments:

* Line wraps should be at 80 characters, and indents two spaces. If you haven't seen it yet [here](https://github.com/numenta/nupic/wiki/Python-Style-Guide#coding-style) is our Python style guide. Your IDE should have settings to do these things automatically.
* Be sure to merge master into your branch.

nab/detectors/gaussian/windowedGaussian_detector.py
subutai: Well, if the Q-function returns 0 then it should definitely be treated as an anomaly!  I just realized that you might want to return 1.0 - tail probability because NAB detects anomalies that are above a threshold.

zuhaagha:
@subutai @BoltzmannBrain Thanks! I've made all the changes. 
I changed the probability to 1 - tail probability and have been tuning the window size and plotting the graphs but it still seems to do poorly. The graphs show that there are way too many false positives and the detector is really sensitive to changes in the streaming data. Maybe tuning more will show an improvement.

subutai:
@zuhaagha Thanks. Does the PR now reflect the optimal windowSize setting? Once that is pushed it's 👍 from me.

zuhaagha:
Not yet! I am still tuning the windowSize and will update it as soon as I get the optimal.

zuhaagha:
@subutai @BoltzmannBrain The highest scores for windowedGaussian detector are, 
reward_low_FP_rate: -27.45 reward_low_FN_rate: 29.71 standard: 14.39
In comparison to random, it only does better on reward_low_FN profile but does quite poorly on reward_low_FP profile. 
I tried a variation as well by using anomaly_likelihood.estimateNormal() method to get parameters but it does not affect the scores that much.

zuhaagha:
The window size for the above scores is 6400 and step size is 100. I tried window sizes between 50 to 8000 at an intervals of 100 or 500.

BoltzmannBrain:
@zuhaagha with these optimal params have you inspected visually? It would be helpful to look at files where the algorithm does poorly and files where it does well. For example, it does well on "realTraffic/speed_t4013.csv", but poorly on all tweet volume files; in fact, it makes almost no detections across all the tweet volume files. My assumption here is these streams are very noisy/spiky and estimating a normal distribution is insufficient, similar to why the Bayesian change point algorithm does poorly.

zuhaagha:
@BoltzmannBrain Yes, I inspected it visually and showed some the results to @subutai yesterday. The detector detects a lot of false positives for data files that are spiky (with flat lines in between) as it treats every spike outside the window as an anomaly. 







The detector also misses true positives and makes no detections in in files like the ones below (where there are no flat lines between spikes and the data fluctuates within a very narrow range or there are spikes but the height of spikes does not vary too much). Another thing I noticed is that the detector sometimes detects a lot of anomalies (false positives) in the probation period (purple window). 









The detector works fine for files like these.

subutai:
@zuhaagha Thanks for the analysis - this will be very helpful. Can you push the final parameters so we can merge?

zuhaagha:
Sure! I will push the parameters and result files.

zuhaagha:
@subutai @BoltzmannBrain Just pushed all results with final parameters

BoltzmannBrain:
Thank you for the plots @zuhaagha

> it treats every spike outside the window as an anomaly. 

[The plotting script only plots the first (the scored) TP detection in the windows](https://github.com/numenta/NAB/blob/master/scripts/plot.py#L340), so I assume the detector is also flagging the spikes in the windows.

> sometimes detects a lot of anomalies (false positives) in the probation period (purple window)

This is expected. The others do the same. The algorithms need some time to learn the data stream. In this case the algorithm is estimating from a prior distribution before that prior is accumulated.

> The detector works fine for files like these. 

There are no anomalies in this file, so it's not the best example of the detector doing well.

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,84 @@
...
+    The anomalyScore is the tail probability of the gaussian (normal) distribution
+    over a sliding window of inputData values. The tail probability is based on the
+    Q-function. The windowSize has been tuned to give best performance on NAB.
+    """
+
+    anomalyScore = 0.0
+    inputValue = inputData["value"]
+    if len(self.windowData) > 0:
+      anomalyScore = 1 - anomaly_likelihood.normalProbability(inputValue,
+                                                              {"mean": self.mean,"stdev": self.std})
```
BoltzmannBrain: This formatting is odd. Better would be

```
anomalyScore = 1 - anomaly_likelihood.normalProbability(
		inputValue,{"mean": self.mean, "stdev": self.std})
```

nab/detectors/gaussian/windowedGaussian_detector.py
```diff
@@ -0,0 +1,84 @@
...
+  def _updateWindow(self):
+    self.mean = np.mean(self.windowData)
+    self.std = np.std(self.windowData)
+    if self.std == 0.0:
+      self.std = 0.000001
+
+
+
+
+
```
BoltzmannBrain: Please use one blank line at the end of files.

BoltzmannBrain:
I left some comments on formatting, otherwise 👍

zuhaagha:
Thanks @BoltzmannBrain. That makes sense
> There are no anomalies in this file, so it's not the best example of the detector doing well.
I was thinking of the detector doing well in terms of no false positives. But yes, there are instances of it doing well like the ones below,

subutai:
All good examples. I imagine the set of files where this does well is very similar to the set of files where Bayesian change point detection does well.

PR #252, opened 2016-06-20
Fix misc. bugs with Gaussian detector, update thresholds json

BoltzmannBrain:
FYI @zuhaagha


zuhaagha:
Thank you for the fix @BoltzmannBrain

PR #261, opened 2016-07-21
Update numenta results following nupic #3195

BoltzmannBrain:
Update the results for the Numenta and NumentaTM detectors, fixing the NuPIC regression tests: https://jira.numenta.com/browse/NUP-2297


BoltzmannBrain:
@subutai merging so we can test the nupic regression tests

PR #278, opened 2016-10-07
Change way processes are added to multiprocessing pool so the pool re…

scottpurdy:
…spects keyboard interrupts.

@oxtopus please review

Fix discovered here:
http://stackoverflow.com/a/1408476

### DIFF
--- a/nab/runner.py
+++ b/nab/runner.py
@@ -129,6 +129,8 @@ class Runner(object):
 
           count += 1
 
+    # Using `map_async` instead of `map` so interrupts are properly handled.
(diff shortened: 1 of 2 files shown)


scottpurdy:
It is a major issue as is leaves a bunch of processes around that you have to manually kill. And I thought this was the cleanest solution. FWIW, `map` actually calls `map_async` under the hood so I don't think this is hacky, it's just moving to the lower level API to avoid the bug.

oxtopus:
LGTM provided a comment is added explaining the choice of `map_async` + timeout vs `map` alone.

PR #288, opened 2016-11-29
Don’t calculate likelihood for spatial anomalies

lscheinkman:
Adresses @cogmission comment https://github.com/numenta/NAB/pull/237#issuecomment-263470002

@scottpurdy The same optimization could be applied to the [numenta_detector.py](https://github.com/numenta/NAB/blob/master/nab/detectors/numenta/numenta_detector.py#L103)

### DIFF
--- a/nab/detectors/htmjava/htmjava_detector.py
+++ b/nab/detectors/htmjava/htmjava_detector.py
@@ -89,11 +89,12 @@ class HtmjavaDetector(AnomalyDetector):
       self.minVal = value
 
     # Compute log(anomaly likelihood)
-    anomalyScore = self.anomalyLikelihood.anomalyProbability(
-      inputData["value"], rawScore, inputData["timestamp"])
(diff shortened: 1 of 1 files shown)


nab/detectors/htmjava/htmjava_detector.py
```diff
@@ -89,11 +89,12 @@ def handleRecord(self, inputData):
       self.minVal = value
 
     # Compute log(anomaly likelihood)
-    anomalyScore = self.anomalyLikelihood.anomalyProbability(
-      inputData["value"], rawScore, inputData["timestamp"])
-    logScore = self.anomalyLikelihood.computeLogLikelihood(anomalyScore)
     if spatialAnomaly:
       logScore = 1.0
+    else:
```
subutai: @lscheinkman The call to anomalyProbability updates internal estimates of the anomaly likelihood. We should call it even in the case of spatial anomalies.

lscheinkman: Got it.

PR #293, opened 2017-03-01
Add --optimize to README since it is necessary when running NAB to ge…

scottpurdy:
…t correct scores.


scottpurdy:
@oxtopus - the following paragraph doesn't explain any of the "action" options currently, just that it "runs" NAB only for the Numenta detector (`-d numenta`). It does seem like some more explanation of the command line script "actions" (detect, optimize, score, normalize) would be useful. Can we do that as follow up work?

Created #294

PR #301, opened 2017-04-26
RES-517: explicit dependency on NuPIC

marionleborgne:
Fixes https://jira.numenta.com/browse/RES-517

### DIFF
--- a/nab/detectors/numenta/numentaTM_detector.py
+++ b/nab/detectors/numenta/numentaTM_detector.py
@@ -25,7 +25,10 @@ import simplejson as json
 from nupic.algorithms import anomaly_likelihood
 from nupic.frameworks.opf.common_models.cluster_params import (
   getScalarMetricWithTimeOfDayAnomalyParams)
-from nupic.frameworks.opf.modelfactory import ModelFactory
+try:
+  from nupic.frameworks.opf.modelfactory import ModelFactory
+except:  # Try importing it the old way (version < 0.7.0.dev0)
+  from nupic.frameworks.opf.model_factory import ModelFactory
 
 from nab.detectors.numenta.numenta_detector import NumentaDetector
 
--- a/nab/detectors/numenta/numentaTM_detector.py
+++ b/nab/detectors/numenta/numentaTM_detector.py
@@ -26,9 +26,10 @@ from nupic.algorithms import anomaly_likelihood
 from nupic.frameworks.opf.common_models.cluster_params import (
   getScalarMetricWithTimeOfDayAnomalyParams)
 try:
-  from nupic.frameworks.opf.modelfactory import ModelFactory
-except:  # Try importing it the old way (version < 0.7.0.dev0)
   from nupic.frameworks.opf.model_factory import ModelFactory
+except:
+  # Try importing it the old way (version < 0.7.0.dev0)
+  from nupic.frameworks.opf.modelfactory import ModelFactory
 
 from nab.detectors.numenta.numenta_detector import NumentaDetector
 
--- a/README.md
+++ b/README.md
@@ -136,12 +136,25 @@ This will install the required modules.
 
 Recommended:
 
-	python setup.py install --user
+	pip install . --user
 
-Or if you are actively working on the code and are familiar with manual
+
+> Note: If NuPIC is not already installed, the version specified in 
+`NAB/requirements.txt` will be installed. If NuPIC is already installed, it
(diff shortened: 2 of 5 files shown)


rhyolight:
I do  not understand this error: https://travis-ci.org/numenta/NAB/builds/226170436

I'm going to re-run the build.

scottpurdy:
Hold off on merge, looking into something

scottpurdy:
This PR captures the JIRA and what we had discussed correctly. I remembered after looking at the NAB code that we originally intended to not require nupic to be installed. I think it is fine for it to be installed by default, but it shouldn't be required. And we also probably don't want to install nupic if users already have a version installed. I.e. if I do a source install of nupic from master and then install nab, I don't want it to uninstall nupic and install and older version.

So I propose that we extend this PR with the following:

- Mimic the code in nupic for allowing developer installs of nupic.bindings [here](https://github.com/numenta/nupic/blob/23d32924eb9a2392649e3869546699648cddbc49/setup.py#L46-L70). But instead, put this here in NAB and allow developer installs of nupic.
- Update the documentation to communicate that existing nupic installs (if later than 0.5.6) will be left alone but if nupic isn't installed then the target version (0.5.6) will be installed
- Change the example commands to use `pip` instead of directly running `setup.py`
- Also include in the documentation instructions for skipping dependencies (`pip install --no-deps .`) if users want to manage the versions themselves

I'm happy to help with this if it makes sense to you guys (@rhyolight @marionleborgne )

marionleborgne:
@scottpurdy This makes sense to me. I'll add these changes.

marionleborgne:
@scottpurdy I addressed you feedback in https://github.com/numenta/NAB/pull/301#issuecomment-297548039

[x] Mimic the code in nupic for allowing developer installs of nupic.bindings [here](https://github.com/numenta/nupic/blob/23d32924eb9a2392649e3869546699648cddbc49/setup.py#L46-L70). But instead, put this here in NAB and allow developer installs of nupic.
[x] Update the documentation to communicate that existing nupic installs will be left alone but if nupic isn't installed then the target version will be installed
[x] Change the example commands to use pip instead of directly running setup.py
[x] Also include in the documentation instructions for skipping dependencies (pip install --no-deps .) if users want to manage the versions themselves


Questions:
1. Shouldn't we have `nupic` in requirements.txt instead of `nupic==0.5.6`?
2. I added a check to determine if nupic is installed or not based on [your code example](https://github.com/numenta/nupic/blob/23d32924eb9a2392649e3869546699648cddbc49/setup.py#L46-L70) but in my case this type of logic does find that nupic is installed. I have it installed from source. Any thoughts?

rhyolight:
There still needs to be a conditional import for [this](https://github.com/numenta/NAB/blob/1bb58d4a68ff457f4e37a6e31c1ea89a8b08fd41/nab/detectors/numenta/numenta_detector.py#L26) and [this](https://github.com/numenta/NAB/blob/1bb58d4a68ff457f4e37a6e31c1ea89a8b08fd41/nab/detectors/numenta/numentaTM_detector.py#L28). Something like

```python
try:
  from nupic.frameworks.opf.modelfactory import ModelFactory
except:
  from nupic.frameworks.opf.model_factory import ModelFactory
```

marionleborgne:
Addressed your feedback @rhyolight

nab/detectors/numenta/numentaTM_detector.py
```diff
@@ -25,7 +25,10 @@
 from nupic.algorithms import anomaly_likelihood
 from nupic.frameworks.opf.common_models.cluster_params import (
   getScalarMetricWithTimeOfDayAnomalyParams)
-from nupic.frameworks.opf.modelfactory import ModelFactory
+try:
+  from nupic.frameworks.opf.modelfactory import ModelFactory
+except:  # Try importing it the old way (version < 0.7.0.dev0)
```
rhyolight: `modelfactory` is the old way, `model_factory` is the new way.
Also, [avoid inline comments](http://nupic.docs.numenta.org/contributing/python-style-guide.html#inline-comments).

rhyolight:
@marionleborgne back to you ;)

marionleborgne:
@rhyolight Gotcha! Fixed :ping_pong:

rhyolight:
@scottpurdy Do you want the last review?

scottpurdy:
Thanks Marion for putting in the extra effort on this - looks great now!

PR #328, opened 2019-01-06
Sweep scoring

balladeer:
Implements a method to score and optimize a detector's results on the NAB corpus by 'sweeping' through an anomaly-sorted list of data points, progressively scoring more and more points as the threshold for what constitutes an 'active' prediction is lowered.

Running locally in a virtual machine, I can score an entire detector (full corpus, all profiles) in about 20 seconds.

Some thoughts:
- All of the unit tests pass locally.  I've added some for the new code and updated existing unit tests.
- I'm not attached to the 'Sweeper' name at all, but I needed _something_ to use, given that the same core code is used in both the optimization and scoring steps.
- I've compared this method vs. the results stored in the `results/` folder.  Any differences in final score tended to be small, and seem to be accompanied by slight updates to the 'best' threshold, as well.  I _think_ that these deviations are expected, but I'd love some confirmation.
- I have _not_ added the ability to save the 'score vs threshold' data to files.  Doing so may help folks evaluate how sensitive various detectors are to changes in threshold, but I wanted to keep changes to a minimum as much as possible. There's already a decent amount of code change here!


rhyolight:
I will review this and run NAB locally to ensure everything works.

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+
+  def calcScoreByThreshold(self, anomalyList):
+    scorableList = self._prepAnomalyListForScoring(anomalyList)
+    scoreParts = self._prepareScoreByThresholdParts(scorableList)
+    scoresByThreshold = []  # The final list we return
+    curThreshold = 1.1  # Start threshold eliminates all rows --> full false negative score
+
+    # Initialize counts:
+    # * every point in a window is a false negative
+    # * every point outside a window is a false positive
```
balladeer: this should say 'every point outside of a window is a true negative'

earthgecko:
I have done some testing and it reduces my run times on my new detector tests from ~30 mins to 7mins and it produces the same scores.  It is almost unbelievable to see
```
  Running optimize step
  Optimizer found a max score of -68.4329257075 with anomaly threshold 1.0.
  Optimizer found a max score of -44.2918503 with anomaly threshold 1.0.
  Optimizer found a max score of -13.2918503 with anomaly threshold 1.0.
```
run in mere seconds! Fantastic @balladeer :1st_place_medal:

rhyolight:
Running this now...

rhyolight:
Vanilla scoring took `10:59.29` on my 20-core iMac Pro. Part of the CPU load looked like this:

screen shot 2019-01-10 at 10 51 29 am

Sweep scoring took `3:48.92` and looked like this:

screen shot 2019-01-10 at 10 51 29 am copy

Final sweep scores fall within range of reported NuPIC scores on the current leaderboard except for the `standard` profile:

```
'reward_low_FP_rate' profile = 61.71
'reward_low_FN_rate' profile = 74.32
'standard' profile = 69.67
```

Leaderboard currently looks like this:

screen shot 2019-01-10 at 11 11 56 am

As you can see `69.67` falls slightly below `69.7` on the leaderboard. Is this a concern @subutai or @lscheinkman?

If not, I will continue to review the code in the PR.

lscheinkman:
@rhyolight I would not worry about the rounding from `69.67` to `69.7`, these are considered the same score as far as the Leaderboard

earthgecko:
@rhyolight from what I can tell the scores on the scoreboard are always rounded from the scores reported by NAB so 69.67 becomes 69.7 on the scoreboard.  However I am not sure that helps if you are trying to determine if the **actual** NAB score changed, surely someone has a log :)

subutai:
Agreed, I wouldn't worry about the rounding.

rhyolight:
Ok great, I am reading the NAB white paper so I can do an informed review. 

FYI if this change is merged it will require a Major version change (because it changes the scoring mechanism), so we’ll need to be extra diligent.

subutai:
> FYI if this change is merged it will require a Major version change (because it changes the scoring mechanism), so we’ll need to be extra diligent.

Is this just a change to the optimizer, or the scoring algorithm itself?  I'm ok with changing the former, but not the latter.  The optimizer is in some sense a convenience function, and it would be awesome if it was a lot faster.

subutai:
> Is this just a change to the optimizer, or the scoring algorithm itself? 

A simple way to test this is to run NAB using the original thresholds file, without doing detection and without running the optimizer. (This should go fast since the algorithms won't be rerun). If we still get the same scores for all the algorithms, we should be good. But make sure you run on all the original  files except for the changes in this PR.

balladeer:
Wow, thanks everyone for taking a look at this PR!

> Is this just a change to the optimizer, or the scoring algorithm itself?

> A simple way to test this is to run NAB using the original thresholds file, without doing detection and without running the optimizer.

While I did change the specific implementation of the scoring algorithm, I did my best to keep the underlying algorithm the same (according to both the whitepaper and the existing implementation).  Running the sweep scoring on the original threshold files, I get the following differences:

| detector  |      profile       | decimals_identical |   original_score   |    sweep_score    |
|-----------|--------------------|--------------------|--------------------|-------------------|
| knncad    | reward_low_FN_rate |                 11 |  64.81232086001505 | 64.81232086001437 |
| knncad    | reward_low_fp_rate |                 10 |  43.40851703291233 | 43.40851703293104 |
| knncad    | standard           |                  8 |  57.99434335898808 | 57.99434335900862 |
| numentaTM | reward_low_FN_rate |                 13 | 69.185068229060349 | 69.18506822906035 |
| numentaTM | standard           |                 13 | 64.553464412543107 | 64.55346441254311 |
| random    | reward_low_FN_rate |                 13 | 25.876351314080456 | 25.87635131408046 |
| random    | reward_low_FP_rate |                 14 |  5.763682319827586 | 5.763682319827587 |

While the differences are few (7 out of 39 detector-profile pairs) and small (differences only show at 9 decimal places and beyond), I certainly leave it to you all as to whether this constitutes a change to the scoring algorithm.

I would love to see someone else's confirmation / results.

-----
> Part of the CPU load looked like this

Out of curiosity, @rhyolight, what did you use to generate those CPU profiles?

rhyolight:
> Out of curiosity, @rhyolight, what did you use to generate those CPU profiles?

The "Activity Monitor" app that comes standard on macOS.

rhyolight:
> run NAB using the original thresholds file, without doing detection and without running the optimizer.

Like this, right?

```
python run.py --score --normalize -t config/thresholds.json
```

rhyolight:
@subutai I have run a comparison of the scoring / normalization processes on master and on this branch. The command I ran was `python run.py --score --normalize`. The default thresholds file was used for each run. Results of each run were written to `results/final_results.json` and had insignificant differences:

```diff
     "numentaTM": {
-        "reward_low_FN_rate": 69.185068229060349,
+        "reward_low_FN_rate": 69.18506822906035,
         "reward_low_FP_rate": 56.665308225043105,
-        "standard": 64.553464412543107
+        "standard": 64.55346441254311
     },
     "random": {
-        "reward_low_FN_rate": 25.876351314080456,
-        "reward_low_FP_rate": 5.763682319827586,
+        "reward_low_FN_rate": 25.87635131408046,
+        "reward_low_FP_rate": 5.763682319827587,
         "standard": 16.831768350517237
     },
```

While the scoring results are essentially the same, I must note that all the result CSV files are entirely different, which is a result of the scoring algorithm being updated. 

As far as optimization, with previous scoring logic the command took `1m22s` to run. The new sweep scoring logic took `23s`. I ran this command (and others) lots of times and this was certainly notable. 

The scoring mechanism is different, but it is achieving nearly the exact same numbers with drastic CPU savings. I am continuing to study the current twiddle algorithm and how this new one works before approving the PR. But from a functional standpoint, this does not need a major version change.

nab/scorer.py
```diff
@@ -18,274 +18,11 @@
...
-                                            within a window calculated per the
-                                            rules above.
-
-  @return (float)
-  """
-  if relativePositionInWindow > 3.0:
-    # FP well behind window
-    return -1.0
-  else:
-    return 2*sigmoid(-5*relativePositionInWindow) - 1.0
```
rhyolight: This is the primary chunk of logic being replaced. The rest of the changes in this file are plumbing. All this logic can now be found in `sweeper.py`. (note to self)

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+                                            within a window calculated per the
+                                            rules above.
+
+  @return (float)
+  """
+  if relativePositionInWindow > 3.0:
+    # FP well behind window
+    return -1.0
+  else:
+    return 2*sigmoid(-5*relativePositionInWindow) - 1.0
```
rhyolight: `sigmoid` and `scaledSigmoid`  are exactly the same.

subutai:
> The scoring code is different, but it is achieving nearly the exact same numbers with drastic CPU savings. I am continuing to study the current twiddle algorithm and how this new one works before approving the PR. But from a functional standpoint, this does not need a major version change.

Sounds great, thank you for the time spent on it! I agree this wouldn't need a major version change.

subutai:
> While I did change the specific implementation of the scoring algorithm, I did my best to keep the underlying algorithm the same (according to both the whitepaper and the existing implementation). 

Looks great - thanks!

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+
+class Sweeper(object):
```
rhyolight: Add class level comment with description of what a sweeper is or what it is for. What is it mean to sweep and why? etc.

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+        curWindowWidth = float(curWindowRightIndex - timestamps.index(curWindowLimits[0]) + 1)
+
+        logger.debug("Entering window: %s (%s)" % (curWindowName, str(curWindowLimits)))
+
+      # If in a window, score as if true positive
+      if curWindowLimits is not None:
+        # Doesn't the `+ 1` in this equation mean we can _never_ have a positionInWindow == 0?
+        positionInWindow = -(curWindowRightIndex - i + 1) / curWindowWidth
+        unweightedScore = scaledSigmoid(positionInWindow)
+        weightedScore = unweightedScore * self.tpWeight / maxTP  # Why is `maxTP` here?
```
rhyolight: Valid question?

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+  def calcSweepScore(self, timestamps, anomalyScores, windowLimits, dataSetName):
+    assert len(timestamps) == len(anomalyScores), "timestamps and anomalyScores should not be different lengths!"
+    timestamps = list(timestamps)
+    windowLimits = list(windowLimits)  # Copy because we mutate this list
+    # The final list of anomaly points returned from this function.
+    # Used for threshold optimization and scoring in other functions.
+    anomalyList = []
+
+    # One-time config variables
+    maxTP = scaledSigmoid(-1.0)
```
rhyolight: Bad variable name... still not sure what "TP" means.

nab/optimizer.py
```diff
@@ -37,115 +39,55 @@ def optimizeThreshold(args):
...
-    optimizedScore, optimizedThreshold)
+  (pool,
+   detectorName,
+   profileName,
+   costMatrix,
+   resultsDetectorDir,
+   resultsCorpus,
+   corpusLabel,
+   probationaryPercent,
+   scoreFlag) = args
```
rhyolight: Does this change the expected arrangement of `args` into `optimizeThreshold()`? If so, the comments should be updated. For example, `tolerance` was an argument for the old twiddle algorithm.

nab/sweeper.py
rhyolight: Also, can't you define this outside this function's scope?

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+
+  def _prepareScoreByThresholdParts(self, inputAnomalyList):
+    scoreParts = {"fp": 0}
+    for row in inputAnomalyList:
+      if row.windowName not in ('probationary', None):
+        scoreParts[row.windowName] = -self.fnWeight
+    return scoreParts
+
+
+  def calcSweepScore(self, timestamps, anomalyScores, windowLimits, dataSetName):
```
rhyolight: Add function comment, include cardinality of anomaly scores. From one data set or many? What does it return?

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+from collections import namedtuple
+import logging
+import math
+
+from nab.corpus import Corpus
```
rhyolight: I don't think you need this import anymore.

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+        curWindowName = "%s|%s" % (dataSetName, curWindowLimits[0])
+        curWindowRightIndex = timestamps.index(curWindowLimits[1])
+        curWindowWidth = float(curWindowRightIndex - timestamps.index(curWindowLimits[0]) + 1)
+
+        logger.debug("Entering window: %s (%s)" % (curWindowName, str(curWindowLimits)))
+
+      # If in a window, score as if true positive
+      if curWindowLimits is not None:
+        # Doesn't the `+ 1` in this equation mean we can _never_ have a positionInWindow == 0?
+        positionInWindow = -(curWindowRightIndex - i + 1) / curWindowWidth
```
rhyolight: See [above](https://github.com/numenta/NAB/pull/328/files#diff-6956a7129fa553c42f4a85415f85924fR137):

```python
curWindowRightIndex = timestamps.index(curWindowLimits[1])
```
Isn't that where the `1` is coming from?

balladeer:
Thanks for all the review so far. I'm headed out for the weekend, but I'll respond to the feedback (linting, comments, etc) when I get back. Cheers!

nab/sweeper.py
rhyolight: I also see now that this variable was named this in previous code, and I think it means "true positive"?

tests/integration/true_positive_test.py
```diff
@@ -125,27 +143,38 @@ def testOnlyScoreFirstTruePositiveWithinWindow(self):
 
     timestamps = generateTimestamps(start, increment, length)
     windows = generateWindows(timestamps, numWindows, windowSize)
-    labels = generateLabels(timestamps, windows)
```
rhyolight: You can remove `generateLabels` from the test helpers now.

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+    else:
+      # Make sure to save the score for the last threshold
+      curScore = sum(scoreParts.values())
+      s = ThresholdScore(curThreshold, curScore, tp, tn, fp, fn, tp + tn + fp + fn)
+      scoresByThreshold.append(s)
+
+    return scoresByThreshold
+
+
+  def scoreDataSet(self, timestamps, anomalyScores, windowLimits, dataSetName, threshold):
```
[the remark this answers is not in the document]
balladeer: Comments added.

nab/sweeper.py
balladeer: Class-level comment added.

nab/optimizer.py
balladeer: Argument comments and order updated.

nab/sweeper.py
balladeer: Function comment added.

tests/integration/true_positive_test.py
balladeer: Helper removed and tests fixed.

nab/sweeper.py
```diff
@@ -0,0 +1,245 @@
...
+        fn -= 1
+      else:
+        fp += 1
+        tn -= 1
+
+      if dataPoint.windowName is None:
+        scoreParts["fp"] += dataPoint.sweepScore
+      else:
+        scoreParts[dataPoint.windowName] = max(scoreParts[dataPoint.windowName], dataPoint.sweepScore)
+    else:
```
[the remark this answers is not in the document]
balladeer: No reason to be in the else; updated the block.

nab/sweeper.py
balladeer: Removed.

nab/sweeper.py
balladeer: Yep, leaving this untouched.

balladeer:
Alright, I've pushed changes that I believe responds to all of the feedback:

1. `sweeper.py`, `scorer.py`, and `optimizer.py` all score 10/10 using NuPIC's pylintrc
2. Quite a few functions / classes now have new or expanded doc strings.

Let me know what else needs changing!

subutai:
👍 Thanks @balladeer - really nice work.

nab/sweeper.py
balladeer: Yeah, makes sense, I think.  I removed the question/comment.

Issue #331, opened 2019-01-15
nab_visualizer handle raw_score


earthgecko:
Changes that appear to have been made between nab v1.0 and the current master in terms of the addition of a raw_score field to the numenta_ and numentaTM csv results files seems to have broken nab_visualizer.html

earthgecko:
PR #332 fixes this.

PR #332, opened 2019-01-15
nab_visualizer_handle_raw_score

earthgecko:
In ref to #331 nab_visualizer_handle_raw_score

Additions were added to fix changes that appear to have been made between nab v1.0 and the current master in terms of the addition of a raw_score field to the numenta results files.  Further to this addition, all the dimensions have been added to the legend, apart from raw_score.  Although these cannot be plotted against additional and relevant y axis, as dygraph does not allow for more that 2 y axis and does not handle displaying dynamic y axis based on the data point selected.  However their addition does at least add them as series on the plot that can be moused over and their value are reported in the legend for allowing visualations of the std, reward_low_FP_rate and   reward_low_FN_rate values.

### DIFF
--- a/scripts/nab_visualizer.html
+++ b/scripts/nab_visualizer.html
@@ -61,6 +61,29 @@ function createDiv(container, id){
   return div
 }
 
+// This addition was added to fix changes that appear to have been made between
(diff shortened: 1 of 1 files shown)


subutai:
👍 

Thanks @earthgecko!

Issue #333, opened 2019-01-18
Disable skyline detector by default


earthgecko:
Disabling the skyline detector by default going forward will decrease the default run.py run time significantly.

The intention here is to not deprecate the skyline detector but make it so that it must be declared in an argument to be run.  This way it can still be used in > v1.0 scoring, but the general community is not subjected to the cost on the time and computation overhead of scoring with this specific detector.  Which has proved the longest running of all the detectors.  As discussed in https://github.com/numenta/NAB/pull/330/

This author shall endeavour to submit scores for the skyline detector on any future versions.

rhyolight:
Good initiative! I'm not sure when we will be ready to push a version but this will be included in it. Thanks @earthgecko.

breznak:
@earthgecko this can be closed as fixed (?)

earthgecko:
Hi @breznak yes #335 closes this.  Done.

PR #335, opened 2019-04-22
Disable skyline detector by default

earthgecko:
This change disables the skyline detector by default with the removal of `'skyline'` from the default detectors argument and the addition of a `--enableSkylineDetector` argument, if this argument is set to `True` then `'skyline` is appended to args.detector.

For Disable skyline detector by default #333

### DIFF
--- a/run.py
+++ b/run.py
@@ -132,11 +132,16 @@ if __name__ == "__main__":
   parser.add_argument("-d", "--detectors",
                     nargs="*",
                     type=str,
-                    default=["null", "numenta", "random", "skyline",
-                             "bayesChangePt", "windowedGaussian", "expose",
-                             "relativeEntropy", "earthgeckoSkyline"],
+                    default=["null", "numenta", "random", "bayesChangePt",
+                             "windowedGaussian", "expose", "relativeEntropy",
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -132,12 +132,17 @@ def main(args):
...
-                    default=["null", "numenta", "random", "skyline",
-                             "bayesChangePt", "windowedGaussian", "expose",
-                             "relativeEntropy", "earthgeckoSkyline"],
+                    default=["null", "numenta", "random", "bayesChangePt",
+                             "windowedGaussian", "expose", "relativeEntropy",
+                             "earthgeckoSkyline"],
                     help="Comma separated list of detector(s) to use, e.g. "
                          "null,numenta")
 
+  parser.add_argument("--enableSkylineDetector",
```
subutai: I don't think you need this extra argument. You can always just manually add 'skyline' to the command line argument for detectors even if it not in the default list.

earthgecko: @subutai a valid point, changes made as requested - https://github.com/numenta/NAB/pull/335/commits/a4310ce086f873eda9a78eb76faa7a2dae4bab4f

PR #340, opened 2019-05-15
Convert NAB to Python 3

iandanforth:
This PR does two major things:

1. Updates NAB to Python 3.
2. Moves NuPIC dependent code out of the repository.

Temporarily NuPIC dependent detectors and the Python 2 support code to run them has been moved to https://github.com/iandanforth/NAB-detectors. It is expected that prior to merging this PR that code will be moved into an official Numenta repository and links to that code in the documentation will be properly updated.


nab/scorer.py
```diff
@@ -102,12 +102,12 @@ def scoreCorpus(threshold, args):
 
   # Using `map_async` instead of `map` so interrupts are properly handled.
   # See: http://stackoverflow.com/a/1408476
-  results = pool.map_async(scoreDataSet, args).get(99999999)
+  results = pool.map_async(scoreDataSet, args).get(999999)
```
rhyolight: does it matter that this number changed?

README.md
```diff
@@ -212,27 +201,28 @@ specific label files (see section below). Please type:
...
+
+In 2019 much of NAB was ported to Python 3, however NuPIC itself was not.
+
+NuPIC dependent detectors are available in the [NAB-detectors]() repository.
+
+Please refer to that repository on how to run any of the following detectors:
+
+    numenta
+    numentaTM
+    htmjava
```
rhyolight: I don't think `htmjava` needed to be moved. It should run from its python 3 launcher script.

nab/scorer.py
iandanforth: Yep there's an overflow error if this value is not reduced.

README.md
iandanforth: The 'htmjava_detector.py' script imports from nupic:

https://github.com/iandanforth/NAB-detectors/blob/master/nab/detectors/htmjava/htmjava_detector.py#L25

Should that be removed?

rhyolight: Subutai and I are discussing this, so don't move this back yet.

README.md
```diff
@@ -212,27 +201,28 @@ specific label files (see section below). Please type:
...
 to see all the options.
 
-Note that to replicate results exactly as in the paper you may need to checkout
-the specific version of NuPIC (and associated nupic.core) that is noted in the
-[Scoreboard](https://github.com/numenta/NAB/wiki/NAB%20Scoreboard):
+##### Run HTM with NAB
+
+In 2019 much of NAB was ported to Python 3, however NuPIC itself was not.
+
+NuPIC dependent detectors are available in the [NAB-detectors]() repository.
```
breznak: TODO please add link

nab/scorer.py
breznak: can you make a comment why such value? I gues it's end of the list?

nab/scorer.py
```diff
@@ -102,12 +102,12 @@ def scoreCorpus(threshold, args):
...
 
   # Total the 6 scoring metrics for all data files
   totals = [None]*3 + [0]*6
   for row in results:
-    for i in xrange(6):
+    for i in range(6):
```
breznak: magic "6", NUMBER_OF_XXX?

README.md
rhyolight: That is too bad, we have to move it then. Good call.

rhyolight:
@breznak

> what would happen with Numenta HTM? I guess it's one of the key functionalities of running this benchmark. Is it possible to run NAB in py3 and NumentaHTM still in py2? In other words, will I be able to run the benchmarks with HTM?

There is already a `nupic` docker container in docker hub, and we are going to create a new NAB container in docker hub to run in Python 2.7. This will be able to run the htm-detectors, which will be in the a repo.

NAB going forward will run in Python 3.

README.md
rhyolight: I am going to add some more detail to this before we merge. I think it should be called `htm-nab-detectors`.

rhyolight:
@iandanforth I have not yet run all the detectors in python 3 yet (only the default ones). I will test out the others later this week.

breznak:
> [HTM docker image for py2.7] docker hub to run in Python 2.7. This will be able to run the htm-detectors, which will be in the a repo.

Yes, this is that we'll be able to run HTM (in python2), my question is how will that mix with NAB (py3)? Will NAB process running in python3 be able to execute htm detector in py2? Or how is this used? Are we supposted to run the htm-detector separately, and NAB only collects (+visualizes etc) the results?

nab/scorer.py
iandanforth: The comment on line 107 explains somewhat do you feel that is insufficient?

nab/scorer.py
iandanforth: Done.

iandanforth:
@rhyolight

 I've run through install and a few detectors (not yet all) under Anaconda. Seems to be working now with a more up to date version of boto3. 

Also added scikit-learn to the requirements.txt. Not sure why this hadn't been done previously, but this way after install you should have everything you need to run all available detectors in the repo without further install steps.

nab/scorer.py
breznak: >  line 107 explains somewhat do you feel that is insufficient? 

that should be enough, I must have missed it while reviewing. Thank you

breznak:
Just a few thoughts about this repository while I was reviewing the changes, 

> Moves NuPIC dependent code out of the repository.

- Would you move all the detectors code from the repo? Having only the NAB infrastructure here. Disadvantage is one can't run a detector from this repo, on the other hand, we mostly care about htm/numenta detector, which needs to be installed from the other repo anyway. 
- On the other hand, if detectors were to stay, it would be nice if htm.java could still be available (as it being Java, should not be affected by py2). I reviewed the code it it depends on
  - nupic/anomaly_likelihood (which you can get from the nupic.cpp repo converted to py3) 
  - nupic.opf....cluster_params.py (should be easy to convert to py3, no other nupic dependencies) 
- independent of the above, most/all of the detectors use the following classes. Would it be worth porting them and including in the infrastructure here?
  - anomaly_likelihood (ported to python3 in nupic.cpp) 
  - nupic.opf.models.cluster_params.py
  - maybe date time encoder date.py (ported in nupic.cpp)

rhyolight:
> Are we supposted to run the htm-detector separately, and NAB only collects (+visualizes etc) the results?

Yes. Unfortunately there is no reliable way to run a python 2 program from a python 3 program.

rhyolight:
> * Would you move all the detectors code from the repo? Having only the NAB infrastructure here. Disadvantage is one can't run a detector from this repo, on the other hand, we mostly care about htm/numenta detector, which needs to be installed from the other repo anyway.

We thought about this quite a bit. In the end, we decided to try to leave the detectors with the scoring implementation and only move out the detectors that cannot be run with Python 3. The reason is because it will be easier for users to run NAB and add new detectors if they are included in the codebase. 



> * it would be nice if htm.java could still be available

Yes it would, but we don't have the manpower to make this happen right now. 

Regarding your 3rd point, I would have the same answer. We want to make sure NAB as a benchmark is still valuable and operational. Anything else would take manpower away from research.

rhyolight:
In addition to the HTM detectors this PR removes, there are two other other detectors in the current library that are not currently attached to the NAB runtime:
- `twitterADVec`: requires R runtime
- `randomCutForest`: requires AWS Kinesis

I did not realize there were detectors that were already detached from the NAB runtime. Now I am wondering... since we already have a precedent for detectors in the NAB codebase that do not run within the NAB python runtime... why move the HTM detectors out at all? Maybe it would be acceptable to simply note that these 5 detectors must be run manually for various reasons: 

- `twitterADVec`: requires R runtime
- `randomCutForest`: requires AWS Kinesis
- `numentaTM` & `numenta`: requires Python 2 runtime
- `htmjava`: requires python 2 runtime for anomaly likelihood

I will bring this up tomorrow with the team.

breznak:
>> Are we supposted to run the htm-detector separately, and NAB only collects (+visualizes etc) the results?

> Yes. Unfortunately there is no reliable way to run a python 2 program from a python 3 program.

unfortunate, but understandable, guess these instructions no longer work 
https://github.com/numenta/NAB#run-htm-with-nab

And we'll be supposed to use 
https://github.com/iandanforth/NAB-detectors#running-a-detector

Just please make sure the code in the NAB-detectors repo is ready (?) before merging the PR, And we won't have tests for NAB+numenta detector (anymore, if we had any)

iandanforth:
Closing this PR as detectors are going to be maintained in the original repository. New PR later today.

PR #341, opened 2019-07-02
Convert NAB to Python 3

iandanforth:
- Like the previous PR this updates NAB to Python 3
 - NuPIC dependent detectors remain in their respective /detectors/ directories with Python 2 support code.
 - The `twitterADVec` detector instructions now has its own directory rather than living in the wiki.

Some documentation polish and final testing remains as well as removing relevant pages from the Wiki to reflect these changes.

PR #347, opened 2019-07-10
New detector naive WIP

### LINKED ISSUE #343: Add "copy" model as baseline
I'm not sure what's the correct name, but `T(n+1) = T(n)` predictor. It's dummy but quite good baseline.

breznak:
WIP 

This is based on #341 

- [x] implement naive predictor
- [ ] how to obtain anomaly score from scalars predicted & current? 

Fixes #343


nab/detectors/naive/naive_detector.py
```diff
@@ -0,0 +1,46 @@
...
+
+  def initialize(self):
+    super().initialize()
+    self.predicted = 0.0 #previous value, last seen
+
+  def handleRecord(self, inputData):
+    """The anomaly score is simply the last seen value"""
+    current = float(inputData["value"])
+    inputData['predicted'] = self.predicted
+    anomalyScore = 1 #FIXME how compute anomaly score from predicted, current?
```
breznak: Looking for help: this is the implementation of naive predictor. 
How do I compute anomaly score from the 2 numbers?

Japillow: You could use something like 1-exp(-nx) where n is a hyperparameter (for ex n = 10) and x = abs(current - predicted)/predicted. Other options could be arctan, 1-n/(x+1), etc

nab/detectors/naive/naive_detector.py
```diff
@@ -25,12 +25,22 @@
...
+class NaiveDetector(AnomalyDetector):
 
 
-class NaiveDetector(AnomalyDetector):
+  def initialize(self):
+    super().initialize()
+    self.predicted = 0.0 #previous value, last seen
 
   def handleRecord(self, inputData):
     """The anomaly score is simply the last seen value"""
```
Japillow: change `anomaly score` to `predicted value`

breznak:
> You could use something like 1-exp(-nx) where n is a hyperparameter (for ex n = 10) and x = abs(current - predicted)/predicted. 

Thank you very much, @Japillow ! I've implemented the function, needed to do small adaptations for numeric stability. Could you have a look, please? 

Updated the doc. 

PS: the results are very bad: 0/0/0, I'm not sure if it's in the implementation, the params of the anomaly function, or just that the "naive" predictor does not work that good on these datasets?

nab/detectors/naive/naive_detector.py
```diff
@@ -18,29 +18,74 @@
...
+    compute anomaly score from 2 scalars
+    """
+    if predicted == 0.0:
+      predicted = EPSILON #avoid division by zero
+
+    # the computation
+    x = abs(current - predicted)/predicted
+    score = 1-math.exp(-self.coef * x)
+
+    # bound to anomaly range (should not happen, but some are over)
```
Japillow: If some are over (or below) there must be an error somewhere, though I can't find it. For which values does this happen ?

breznak: > must be an error somewhere, though I can't find it. For which values does this happen ? 

I'll add debug prints. Couldn't get it properly, because the benchmark runs multithreaded.

breznak:
@Japillow could you please help me finish this PR?

PR #360, opened 2019-11-21
Removed obsolete Python 2 conda environment.

rhyolight:


rhyolight:
That is strange, when I created a conda env from this file, I swear it was Python 2.7. But this file obviously is not.

environment.yml
```diff
@@ -1,20 +0,0 @@
-name: NAB
-channels:
-  - defaults
-  - conda-forge
-
-dependencies:
-  - python=3.6
```
lscheinkman: Are you sure this is for python 2 ?

rhyolight: I have already closed the PR

PR #375, opened 2021-01-22
Update setup.py

gibinavineesh:


subutai:
Looks like this is just adding a comment at the end that is not very informative. Closing.

PR #388, opened 2021-11-11
Corrected window for realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_Netwo…

markNZed:
The data file realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_NetworkIn.csv has two anomalies and 1243 samples. The total amount of window length is 124 samples, so 62 for each anomaly, 31 samples prior to each anomaly and 31 samples after each anomaly.

The labels/combined_windows.json file is not correct for the first anomaly at timestamp "2013-10-10 09:35:00" or line 208 in realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_NetworkIn.csv 

The samples are at every 5 minutes, 31 samples prior is 2hours 35mins or timestamp "2013-10-10 07:00:00" (line 177), 31 samples after is "2013-10-10 12:10:00" (line 239)

The file labels/combined_windows.json has been updated.

This does not impact the inputs to the detectors, nor the outputs of the detectors, therefore I think this change should be included in the current NAB release. It would require rerunning the scoring (not the detectors).


markNZed:
I've replaced this with https://github.com/numenta/NAB/pull/389 so I could move the fix into a branch on my repo

Issue #386, opened 2021-11-03
Is Numenta accepting PR and new algorithm submissions to NAB?


markNZed:
I have an anomaly detector that I would like to submit, is Numenta still accepting new submissions?

subutai:
Hi,

Yes, we are. The submitter will need to sign the [Contributor's License](https://numenta.com/contributor-license) and meet [these requirements](https://github.com/numenta/NAB/wiki/NAB-Contributions-Criteria).

Thanks!

markNZed:
Hi, it makes sense that changes to NAB need a contributors license. It is less clear about the actual detector, it seems strange to grant Numenta commercial rights to the intellectual property of all the detectors submitted. I guess Numenta does not have those rights for several of the detectors included in the scoreboard? Is there a way to submit a new detector with a typical open source license rather than transfer the intellectual property to Numenta? I guess the detector could be in another git repo and only the interface for running that detector included in NAB? 

On a related topic I added an issue for an error in the window labels for one of the timeseries. How do we handle bugs like that - does a new detector run with the window corrected or does it run with the incorrect window (like the other detectors have been run)?

subutai:
Hi, signing a contributor's license is pretty standard practice with open source projects. Ours is actually modeled after the [Apache license](https://www.apache.org/licenses/contributor-agreements.html).  Basically in order for us to re-release any code, we need full rights from the contributor to do so.  The contributor of course still retains their rights to do what they want.  All code checked into NAB has gone through this license.

You can also submit a detector that's residing in another repository.  One example is the [Twitter detector](https://github.com/numenta/NAB/tree/master/nab/detectors/twitter). Only the interface for running it is in NAB.  We'd want clear instructions for installation and running it. We do however require that it's open source, not for example, hidden behind some cloud API.

subutai:
> On a related topic I added an issue for an error in the window labels for one of the timeseries. How do we handle bugs like that - does a new detector run with the window corrected or does it run with the incorrect window (like the other detectors have been run)?

Is that #385? Unfortunately anything that might change the results requires another version of NAB so that results are exactly comparable in papers. NAB is in "maintenance mode" right now, so we can't really create a new version, no matter how minor, at the current time.

markNZed:
> Is that #385? 

[...]

Issue #385, opened 2021-11-01
Erroneous window


markNZed:
The first anomaly in realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_NetworkIn.csv occurs at 2013-10-10 09:35:00 https://github.com/numenta/NAB/blob/2fafea9c37e250edf5281d7f23627dc5e4ebfcac/labels/combined_labels.json#L78 but the window starts at 2013-10-10 10:35:00 https://github.com/numenta/NAB/blob/2fafea9c37e250edf5281d7f23627dc5e4ebfcac/labels/combined_windows.json#L168 The label for the window are not correct in the result files.

markNZed:
I added a PR for this https://github.com/numenta/NAB/pull/389

PR #389, opened 2021-12-08
Corrected window for realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_Netwo…

markNZed:
I'm replacing a previous pull request with this because I've moved the fix to a branch. Here is the text from the previous PR https://github.com/numenta/NAB/pull/388 

The data file realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_NetworkIn.csv has two anomalies and 1243 samples. The total amount of window length is 124 samples, so 62 for each anomaly, 31 samples prior to each anomaly and 31 samples after each anomaly.

The labels/combined_windows.json file is not correct for the first anomaly at timestamp "2013-10-10 09:35:00" or line 208 in realAWSCloudwatch/iio_us-east-1_i-a2eb1cd9_NetworkIn.csv

The samples are at every 5 minutes, 31 samples prior is 2hours 35mins or timestamp "2013-10-10 07:00:00" (line 177), 31 samples after is "2013-10-10 12:10:00" (line 239)

The file labels/combined_windows.json has been updated.

This does not impact the inputs to the detectors, nor the outputs of the detectors, therefore I think this change should be included in the current NAB release. It would require rerunning the scoring (not the detectors).


markNZed:
@lscheinkman thanks for looking into this (and thanks for reviewing ARTime too!). Do you think the scores table should be recalculated based on the new label window? I think this should only require rerunning the scoring step but I did not do that in this PR.

lscheinkman:
@markNZed Thank you for pointing that out, however the label window values were not merged into NAB. The results you reported in the scores table are based on the [original label window](https://github.com/numenta/NAB/blob/master/labels/combined_windows.json#L168-L169).

This PR was closed because https://github.com/numenta/NAB/pull/393 was merged and included this commit fa5fe2c. The label window changes were reverted back to their original values in this commit 450b949 so there is no need to rerun the scoring step.

markNZed:
@lscheinkman what happens with the bug in the original label window? I imagine that should be fixed at some point in the future.

lscheinkman:
You can leave issue #385 open so others are aware.

Changes to the label window may require another version of NAB, but as @subutai mentioned in #386, "_NAB is in maintenance mode right now, so we can't really create a new version ..._".

PR #413, opened 2026-07-27
New #1 detector SENTIS addition request

mahammad-yusifov:
Ladies and Gentlemen, I present you SENTIS!

SENTIS was designed by studying the detectors already benchmarked on NAB and borrowing their best ideas while patching their blind spots. ARTime matches normalized value shapes but is structurally blind to slow drift, so SENTIS's Deviation Index (Channel 2) tracks absolute value trends instead. CAD OSE remembers value-transition sequences but clamps anything outside its initial range, so SENTIS's Spatial Envelope (Channel 1) never clamps. Numenta HTM calibrates every alert against the stream's own recent history rather than a fixed threshold — the same idea drives SENTIS's shared likelihood calibration. earthgecko Skyline avoids assuming a Gaussian distribution, and so does SENTIS: every channel is self-calibrating and non-parametric. Six channels, each targeting a distinct anomaly class — Spatial Envelope, Deviation Index, Sequence Memory, Temporal Context, Flatline Sentinel, and a dual-vigilance Shape Memory inspired by ARTime — are fused into one continuous score.


earthgecko:
Congrats @mahammad-yusifov :1st_place_medal: 

I can confirm that I have reproduced the results from the previous PR #413 - there are changes to the code in this PR which I have not tested, but the scores here in this PR have not changed, so they are probably reproducible.

I have also tested this as a standalone algorithm and **it is legit and fast**, and amazingly only uses the math library!

Congratulations @mahammad-yusifov, a very good idea, consolidating the best parts of nab/detectors to make the best nab/detector :+1:

mahammad-yusifov:
> Congrats @mahammad-yusifov 🥇
> 
> I can confirm that I have reproduced the results from the previous PR #413 - there are changes to the code in this PR which I have not tested, but the scores here in this PR have not changed, so they are probably reproducible.
> 
> I have also tested this as a standalone algorithm and **it is legit and fast**, and amazingly only uses the math library!
> 
> Congratulations @mahammad-yusifov, a very good idea, consolidating the best parts of nab/detectors to make the best nab/detector 👍

Thank you, for your interest, for our conversation through mail.

